PR #14, opened 2015-05-22
make --config required, and check config path

sevko:
```
tilequeue/command.py
        -This will make it obvious that the user needs to supply a
        `--config` path, and produce a human error message when the file
        isn't found (you otherwise get a rather ugly and unenlightening
        trace).
```

### DIFF
--- a/tilequeue/command.py
+++ b/tilequeue/command.py
@@ -41,7 +41,8 @@ import time
 
 def create_command_parser(fn):
     def create_parser_fn(parser):
-        parser.add_argument('--config')
+        parser.add_argument('--config', required=True,
+                            help='The path to the tilequeue config file.')
         parser.set_defaults(func=fn)
         return parser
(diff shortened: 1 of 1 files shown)


sevko:
@rmarianski , it seems like all of the subcommands currently require the `--config` argument. Think we should just make it a required argument to the `tilequeue` command?

rmarianski:
Diff looks good to me. :+1: 

All subcommands do require being pointed to a config file. If you can make it a requirement there, feel free, assuming the invocation for all subcommands remains exactly the same. But I don't think this is important, because IIRC adding this config option is shared amongst all code paths.

sevko:
Nevermind. I thought it'd be an easy change, but preserving the current behavior (where a `--config` argument can be passed to the subcommands) while also allowing it to be passed to `tilequeue`  directly will make things messy. Merging as is.

PR #15, opened 2015-05-27
abstract out SQS and S3

sevko:
This pull requests adds (or improves on already existing) drop-in replacements for S3 and SQS that should make it easier for developers to work on tilequeue locally, since they won't have to bother setting up those AWS services.

  * beef up `tilequeue.queue.file.OutputFileQueue`, an on-disk, file-based alternative to SQS, adding locking, read functionality, etc.
  * make `tilequeue.store.TileDirectory` (renamed from `TileFile`) write files out to a directory.
  * add tests for both of the above
  * turn the config's `aws.s3` and `awd.sqs` properties into top-level `queue` and `store` keys, since both the queue and tile-store are now generic and not tied to those services.
  * some other small changes were involved in changing code that was hardcoded against SQS/S3 (like how we get timestamps from each `CoordMessage`).

I can currently `tilequeue seed` and `tilequeue process` in parallel and generate all of the tiles to an on-disk directory, so it worketh. @rmarianski , @baldur , please look this over when you get a chance.


sevko:
Oh, woah, I broke literally all the tests.
...fixing...

config.yaml.sample
```diff
@@ -1,10 +1,12 @@
+queue:
+  type: file
+  name: queue
+store:
+  type: directory
+  name: tiles
+  path: osm
+  reduced-redundancy: true
```
rmarianski: Can we have the sample config default to the types that we use in production, and have a comment for all the other types? Or maybe a separate commented line for each type?

tests/queue/test_file.py
```diff
@@ -0,0 +1,83 @@
...
+    def setUp(self):
+        self.test_tile_coords = [
+            (0, 0, 0),
+            (1, 2, 3),
+            (4, 5, 6),
+            (9, 3, 1),
+            (4, 7, 1)
+        ]
+        self.test_tile_objs = [
+            Coordinate(*coords) for coords in self.test_tile_coords]
```
rmarianski: I think it's better to specify the column, row, zoom explicitly via the keyword properties, and this goes for the other instantiations too, if only to make it clear to the reader. The default order that we get from modest maps is imho unexpected, namely row, column, zoom. It doesn't look like it's what you expected either, since there's some invalid coordinates specified. In this case it doesn't matter though, because there aren't checks for invalid coordinates, but it can still lead to confusion.

tests/queue/test_file.py
```diff
@@ -0,0 +1,83 @@
...
+        self.tile_coords_str = '\n'.join(
+            map(tile.serialize_coord, self.test_tile_objs)) + '\n'
+        self.tiles_fp = StringIO.StringIO()
+        self.queue = OutputFileQueue(self.tiles_fp)
+
+    def test_read(self):
+        self._write_tiles_to_file()
```
rmarianski: A personal preference, but I think it's clearer if _write_tiles_to_file would take in the coords explicitly instead of pulling it from `self`. I think it would be fine if each invocation just took the list of coords to write, which could just be the `self.test_tile_coords` reference.

tests/store.py
```diff
@@ -0,0 +1,55 @@
...
+from tilequeue import store
+from tilequeue import format
+from ModestMaps.Core import Coordinate
+import os
+
+
+class TestTileDirectory(unittest.TestCase):
+
+    def test_write_tile(self):
+        dir_path = 'tests/fixtures/tiles_dir'
```
rmarianski: Can we ask python to give us a temporary path instead?

tilequeue/command.py
```diff
@@ -85,10 +84,22 @@ def make_queue(queue_type, queue_name, cfg):
...
-        # useful for testing
         from tilequeue.queue import OutputFileQueue
-        fp = open(queue_name, 'w')
+        if os.path.exists(queue_name):
+            assert os.path.isfile(queue_name), \
+                'Could not create file queue. `./{}` is not a file!'.format(
+                    queue_name)
+        else:
+            # Create the file.
+            with open(queue_name, 'w'):
```
rmarianski: Is creating an empty file necessary here? Or would the open in append mode take care of that implicitly?

tilequeue/command.py
```diff
@@ -150,7 +161,7 @@ def make_seed_tile_generator(cfg):
 
 
 def tilequeue_drain(cfg, peripherals):
-    queue = make_queue(cfg.queue_type, cfg.queue_name, cfg)
+    queue = peripherals.queue
```
rmarianski: Good catch!

tilequeue/command.py
```diff
@@ -465,7 +489,9 @@ def stop_all_workers(signum, stack):
         for thread_s3_storage_stop in threads_s3_storage_stop:
             thread_s3_storage_stop.set()
         thread_sqs_writer_stop.set()
-        queue_printer_thread_stop.set()
+
+        if queue_printer_thread_stop:
+            queue_printer_thread_stop.set()
```
rmarianski: Good catch here too!

rmarianski:
This looks good. :zap: I put in some comments on what I noticed, which we can discuss further.

Let's be careful around merging this into master. I'd like to have the necessary deployment changes surrounding the config file ready before we hit the green button.

sevko:
I addressed all your comments in a number of follow-up commits.

Before we move forward, did you try running tilequeue off this branch against SQS and S3? I've only been able to test if with the on-disk alternatives that this PR introduces, so I don't know if I mistakenly broke anything on the SQS/S3 side (I don't have access to those, so can't use them).

PR #17, opened 2015-06-04
add option to seed custom bounding-boxes and filter metro-extracts

sevko:
Increase seeding configurability by:
  * allowing users to filter metro-extracts by city name
  * adding the ability to specify any number of custom bounding boxes

### DIFF
--- a/config.yaml.sample
+++ b/config.yaml.sample
@@ -21,6 +21,11 @@ tiles:
       url: https://raw.githubusercontent.com/mapzen/metroextractor-cities/master/cities.json
       zoom-start: 11
       zoom-until: 15
+
+      # Can be set to an array of metro-extract city names, like
+      # 'new-york_new-york', to only seed tiles from those areas rather than
+      # all of them.
+      cities: Null
     top-tiles:
       url: https://gist.github.com/brunosan/b7ce3df8b48229a61b5b/raw/37e42e77f253bc204076111c92acc4d5e653edd2/top_50k_tiles.csv
       zoom-start: 11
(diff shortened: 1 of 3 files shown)


sevko:
@rmarianski , check this out when you get a chance.

config.yaml.sample
```diff
@@ -21,10 +21,24 @@ tiles:
       url: https://raw.githubusercontent.com/mapzen/metroextractor-cities/master/cities.json
       zoom-start: 11
       zoom-until: 15
+
+      # Can be set to an array of metro-extract city names, like
+      # 'new-york_new-york', to only seed tiles from those areas rather than
+      # all of them.
+      cities: Null
```
rmarianski: Can we make this something that is valid yaml? Either: `cities:`, `cities: null`, or `cities: [new-york_new-york]`.

tilequeue/command.py
```diff
@@ -150,7 +157,12 @@ def make_seed_tile_generator(cfg):
     else:
         top_tiles = ()
 
-    combined_tiles = chain(all_tiles, metro_extract_tiles, top_tiles)
+    custom_tiles = tile_generator_for_multiple_bounds(
```
rmarianski: Can this be guarded with an explicit check like the others? Set it to an empty list/tuple if the bounds and/or zooms aren't configured.

tilequeue/config.py
```diff
@@ -102,13 +117,19 @@ def default_yml_config():
...
                     'zoom-start': None,
                     'zoom-until': None,
                 },
-                'should-add-to-tiles-of-interest': True,
+                'custom': {
+                    'zoom-start': None,
+                    'zoom-until': None,
+                    'bboxes': []
+                },
+                'should-add-to-tiles-of-interest': True
```
rmarianski: Good catch!

sevko:
@rmarianski , all addressed.

rmarianski:
:+1:

PR #22, opened 2015-09-02
Add framework to post-process layers

zerebubuth:
This adds a post-processing step which comes after the functions for transforming and sorting the individual layers. The post-processing step has access to the whole tile, and so it can modify existing layers or create new layers based on existing layers. This allows us to "intercut" one layer with another, see https://github.com/mapzen/TileStache/pull/37, or create layers based purely on already-queried data (e.g: centroids).

Note: this is part of trying to solve https://github.com/mapzen/vector-datasource/issues/136.

### DIFF
--- a/tilequeue/process.py
+++ b/tilequeue/process.py
@@ -55,6 +55,13 @@ def _preprocess_data(feature_layers, shape_padded_bounds):
 
     return preproc_feature_layers
 
+# post-process all the layers simulataneously, which allows new
+# layers to be created from processing existing ones (e.g: for
+# computed centroids) or modifying layers based on the contents
+# of other layers (e.g: projecting attributes, deleting hidden
+# features, etc...)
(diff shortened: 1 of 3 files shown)


tilequeue/process.py
```diff
@@ -56,6 +56,29 @@ def _preprocess_data(feature_layers, shape_padded_bounds):
...
+# of other layers (e.g: projecting attributes, deleting hidden
+# features, etc...)
+def _postprocess_data(feature_layers, post_process_data):
+
+    for step in post_process_data:
+        fn = loadClassPath(step['fn_name'])
+        params = step['params']
+
+        layer = fn(feature_layers, **params)
+        if layer is not None:
```
rmarianski: Now that the intercut function won't return None, should we eliminate this check? Or do you foresee other processing functions that may want to signal failures by returning None, and we should keep this around?

PR #25, opened 2015-09-10
Allow post-processing to append new layers.

zerebubuth:
Fix bug where post-processing would only replace existing layers, not add new ones, which we need for mapzen/vector-datasource#140.


rmarianski:
:+1:

zerebubuth:
Thanks!

PR #26, opened 2015-09-10
Simplify geometries during processing step

rmarianski:
Refs #20

PR #27, opened 2015-09-10
Filter out geometries with a small area

rmarianski:
Refs mapzen/vector-datasource#151


rmarianski:
This is based on the branch from #26. 

@zerebubuth review?

PR #28, opened 2015-09-14
Pass zoom parameter to the post-processing functions.

zerebubuth:
We want to not compute water boundaries for mapzen/vector-datasource#203, so the post-processor needs to know what the zoom is to switch itself off.


rmarianski:
:+1:

PR #33, opened 2015-10-13
Added bbox_intersection filter for Jinja.

zerebubuth:
Added bbox_intersection filter for Jinja, which allows clipping to the query bounding box. Although doing an intersection operation in the database rather than on the client can take some time, it seems it's worthwhile in this case to avoid sending too much over the network.

Connects to mapzen/vector-datasource#302.


rmarianski:
:+1:

PR #36, opened 2015-10-22
Add support to load WOF neighbourhoods

rmarianski:
Refs https://github.com/mapzen/vector-datasource/issues/147. Related to https://github.com/mapzen/vector-datasource/pull/325.

I spelled it "neighbourhoods" throughout, because it seemed like it would be more consistent with OSM and the WOF data itself.

To make the update atomic, I truncate the existing table (if it has existing rows) and insert the new data via `COPY FROM` in a transaction. It might have been better to instead create a new table separately, and then only drop the old table and rename the new in a transaction. But it was easier to implement this way. Locally that transaction takes about 0.5 seconds, which is imho acceptable for a weekly or even daily update.

I also created a new abstraction to handle enqueueing sqs messages via threads. A few commands had ad-hoc approaches to that, and I figured it would be useful to try and normalize that a little bit and remove some duplication.


zerebubuth:
:+1: looks good to me. My only comment is that we seem to be hard-coding `n_threads` in a couple of places, but both of them seem to be I/O bound, so that's not a big deal.

rmarianski:
@zerebubuth good point. I added the ability to make it configurable in: https://github.com/mapzen/tilequeue/commit/d95af001b596cfa2005d04f354eb1700d47b3cdc.

I also realized that we could re-use the enqueuer abstraction when performing the intersection process too; I missed that earlier. That's added in: https://github.com/mapzen/tilequeue/commit/891a4a78e557e842b32a20670fd1bbeea56ee8a0

zerebubuth:
Nice! :+1:

PR #37, opened 2015-10-23
Only update the wof data when changes exist

rmarianski:
Connects to https://github.com/mapzen/vector-datasource/issues/147


zerebubuth:
:+1: looks good to me.

PR #40, opened 2015-10-26
Re-check polygon validity after call to buffer, just in case.

zerebubuth:
Connects to mapzen/TileStache#79.

### DIFF
--- a/tilequeue/process.py
+++ b/tilequeue/process.py
@@ -58,7 +58,7 @@ def _preprocess_data(feature_layers, shape_padded_bounds):
     return preproc_feature_layers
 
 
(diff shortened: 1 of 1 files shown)


tilequeue/process.py
```diff
@@ -58,7 +58,7 @@ def _preprocess_data(feature_layers, shape_padded_bounds):
     return preproc_feature_layers
 
 
-# post-process all the layers simulataneously, which allows new
+# post-process all the layers simultaneously, which allows new
```
rmarianski: Nice catch!

rmarianski:
:+1:

PR #41, opened 2015-10-27
Parse WOF neighbourhoods using header names

rmarianski:
The assumption here is that the header indexes would change more often than the names. This should help keep the parsing logic more resilient in the case when a new column is added.


zerebubuth:
Nice future-proofing! :+1:

PR #42, opened 2015-11-02
Update WOF processing pipelines

rmarianski:
Depends on https://github.com/mapzen/vector-datasource/pull/365

* Add wof-load-initial-neighbourhoods command

This will be run initially to load up the database the first time. It
requires a local checkout of the WOF data repository on the filesystem.

* Update wof-process-neighbourhoods command

Instead of truncating everything and re-inserting, start tracking the
additions, modifications, and removals and execute the necessary sql to
sync neighbourhoods.

### DIFF
--- a/tilequeue/wof.py
+++ b/tilequeue/wof.py
@@ -1,16 +1,20 @@
 from collections import namedtuple
 from contextlib import closing
 from cStringIO import StringIO
+from itertools import imap
 from operator import attrgetter
 from shapely import geos
 from tilequeue.tile import coord_marshall_int
 from tilequeue.tile import coord_unmarshall_int
 from tilequeue.tile import mercator_point_to_coord
 import csv
+import json
+import os.path
 import psycopg2
 import pyproj
 import Queue
 import requests
+import shapely.geometry
 import shapely.wkb
 import threading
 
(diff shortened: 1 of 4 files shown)


tilequeue/wof.py
```diff
@@ -66,33 +75,194 @@ def fetch_wof_neighbourhoods(url):
...
-        neighbourhood = Neighbourhood(wof_id, name, x, y)
-        yield neighbourhood
+        neighbourhood_meta = NeighbourhoodMeta(
+            wof_id, name, label_position, file_hash)
+        yield neighbourhood_meta
 
 
-class WofNeighbourhoodFetcher(object):
+def fetch_wof_url_meta_neighbourhoods(url):
+    r = requests.get(url, stream=True)
```
zerebubuth: Do we want to add some retry logic here, or does `requests` handle that?

rmarianski: We probably do, yes. And we probably need to catch exceptions in case of unexpected failures, and check if we actually got a 200 back. Not sure what's best to do in these failure cases. Maybe have the caller log an error and bail on the entire run?

zerebubuth: Yeah, might be worth adding `assert r.status_code == 200, 'Failure requesting: %s' % url` like the other location. Not sure how the retry would look, or whether it's worth retrying in both places. I mean, we can't do anything unless we've got the metadata, but I'm not sure how we'd work around the 503 issue without introducing a lot of extra complexity.

rmarianski: Added the assertion in 0ad66eb.

Looks like `requests` does have support for [retrying](http://docs.python-requests.org/en/latest/api/#requests.adapters.HTTPAdapter) out of the box, but this wouldn't help with the 503s. One idea is that we could try creating a custom http adapter that would consider 503 responses as a signal to sleep and retry again.

But maybe it's better to just punt on this for now and wait if we see this kind of problem in practice?

zerebubuth: Fair enough - it's a one-off process anyway, so fingers crossed we can do it without hitting any 503s.

zerebubuth:
:+1:

PR #45, opened 2015-11-03
Track mercator area in separate column

rmarianski:
Connects with: https://github.com/mapzen/vector-datasource/pull/379


zerebubuth:
:+1: looks good to me. Was initially worried about the `int()` conversion, but it looks like not only does `int(x)` return a `long` when `x > sys.maxint` but also `sys.maxint` on a 64-bit system is 2^63-1, so unless there's a neighbourhood which covers the whole world 5,743 times, we're probably okay :wink:.

rmarianski:
Good idea to check anyway. The overflow possibility slipped my mind.

PR #47, opened 2015-11-05
Add resilience to WOF imports

rmarianski:
Connects with: https://github.com/mapzen/vector-datasource/issues/393

This isn't finished yet, but the basic structure of how to handle failures should be in place.

@zerebubuth: could you take a quick look and let me know if I went down the wrong path, or if you have any suggestions thus far?

### DIFF
--- a/tilequeue/wof.py
+++ b/tilequeue/wof.py
@@ -117,36 +117,110 @@ def fetch_wof_url_meta_neighbourhoods(url, placetype):
     return parse_neighbourhood_meta_csv(csv_line_generator, placetype)
 
 
+class NeighbourhoodFailure(object):
+
+    def __init__(self, wof_id, reason, message, halt=False, skipped=False,
+                 funky=False):
+        # halt is a signal that threads should stop fetching. This
+        # would happen during a network IO error or when we get an
+        # unexpected http response when fetching raw json files. In
+        # some scenarios this could be recoverable, but because that
+        # isn't always the case we assume that we should stop further
+        # requests for more raw json files, and just process what we
+        # have so far.
+
+        # skipped means that we won't log this failure, ie there was
+        # an earlier "halt" error and processing of further records
+        # has stopped.
+
+        # funky is a signal downstream that this is a "soft" or
+        # expected failure, in the sense that it only means that we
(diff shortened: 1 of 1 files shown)


tilequeue/wof.py
```diff
@@ -689,30 +799,38 @@ def fetch_toi():
...
+            failed_wof_ids.add(failure_wof_id)
+            ids_to_add.discard(failure_wof_id)
+            ids_to_update.discard(failure_wof_id)
+
+        # now we'll want to ensure that these ids are not present in the diffs
+        new_diffs = []
+        for n1, n2 in diffs:
+            if n2 is not None and n2.wof_id in failed_wof_ids:
+                continue
+            new_diffs.append((n1, n2))
```
zerebubuth: Might this be easier to understand if it were?

```python
if n2 is None or n2.wof_id not in failed_wof_ids:
    new_diffs.append((n1, n2))
```

tilequeue/wof.py
```diff
@@ -689,30 +799,38 @@ def fetch_toi():
...
-        # once we detect fetch failures, and I think it's best to
-        # handle all these cases at once, but just making a note in
-        # here now to serve as a reminder
+        # we should just remove any neighbourhoods from add/update lists
+        # also keep track of these ids to remove from the diffs too
+        failed_wof_ids = set()
+        for failure in failures:
+            failure_wof_id = failure.wof_id
+            # TODO logging funky failures is noisy
+            # maybe we just log the count of funky failures?
```
zerebubuth: I would agree - if we're expecting that we might get quite a lot of these in the future, to log them every hour is just going to bloat the logs. Does this mean that `NeighbourhoodFailure` needs a flag for funkiness?

tilequeue/wof.py
rmarianski: Agreed, updated in e771584.

tilequeue/wof.py
rmarianski: Yup, added that and suppressed individual logging for funky records in 3f5de44.

zerebubuth:
:+1: looks good to me

PR #48, opened 2015-11-06
Add handling for superseded_by inconsistencies

rmarianski:
This handles the edge case where the `superseded_by` is set in the json file but not in the meta csv.


zerebubuth:
:+1: looks good to me.

PR #49, opened 2015-11-09
Update failure logging to stay on one line

rmarianski:

### DIFF
--- a/tilequeue/wof.py
+++ b/tilequeue/wof.py
@@ -714,7 +714,7 @@ def generate_tile_expiry_list(zoom, diffs):
 def log_failure(logger, failure):
     if not (failure.skipped or failure.funky or failure.superseded):
         failure_message_one_line = failure.message.replace('\n', ' | ')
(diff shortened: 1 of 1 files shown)


zerebubuth:
Good point! In case of other weird characters, might it be worth quoting the whole thing? Something like:

```python
logger.error('Neighbourhood failure for %s: %s - %s' % (
  failure.wof_id, repr(failure.reason), repr(failure_message)))
```

Then we can see when `reason` or the message might be `None` or have other, non-printable characters in them?

rmarianski:
Good point. The `repr` formatting might give us some extra quotes, but that'll probably be more useful for debugging and safer too. Added in 2d15784.

zerebubuth:
Ooh, `%r` is `repr`? That's cool, I didn't know that!

:+1:

rmarianski:
Yup, `%r` is the equivalent of `repr`. It certainly looks like that locally anyway :)

PR #50, opened 2015-11-09
Escape string values in sql

rmarianski:
@zerebubuth, @nvkelso: could you review please?

### DIFF
--- a/tilequeue/wof.py
+++ b/tilequeue/wof.py
@@ -559,12 +559,15 @@ class WofModel(object):
         def nullable_int(x):
             return 'NULL' if x is None else str(x)
 
+        def escape_string(s):
+            return s.replace("'", "''")
+
         def gen_data(n):
             return dict(
                 table=self.table,
                 placetype=neighbourhood_placetypes_to_int[n.placetype],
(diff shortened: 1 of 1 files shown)


nvkelso:
Does the Postgres library include this for us?

rmarianski:
It doesn't, but I may be able to re-arrange things a little to have it escape the parameters when executing the statement itself. It seems like that's the recommended way to do it. I'll look into that.

rmarianski:
I updated the sql command execution to use named query parameters. As a bonus, it simplified some of the code too.

nvkelso:
Generally looks good. 

Why don't we need to set the SRID anymore?

rmarianski:
We tell shapely to encode it in the wkb: https://github.com/mapzen/tilequeue/pull/50/files#diff-3c589dc895a7fd4ecd09a8c77c9cb336R561

zerebubuth:
:+1: :zap:

Nice! Makes me wish I'd [RTFM](http://initd.org/psycopg/docs/usage.html#query-parameters)ed properly! :wink:

PR #51, opened 2015-11-11
Remove neighbourhoods that became funky

rmarianski:
Refs https://github.com/mapzen/vector-datasource/issues/403

### DIFF
--- a/tilequeue/wof.py
+++ b/tilequeue/wof.py
@@ -849,13 +849,18 @@ class WofProcessor(object):
         # also keep track of these ids to remove from the diffs too
         failed_wof_ids = set()
         superseded_by_wof_ids = set()
-        n_funky = 0
+        funky_wof_ids = set()
         for failure in failures:
             failure_wof_id = failure.wof_id
             log_failure(self.logger, failure)
 
             if failure.funky:
-                n_funky += 1
+                # this scenario is triggered for new neighbourhoods,
+                # or if a neighbourhood became funky
+                # we handle both of these scenarios in tests later on,
+                # but for now we just track the id of the funky
+                # neighbourhoods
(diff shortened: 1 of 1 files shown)


tilequeue/wof.py
```diff
@@ -892,6 +898,14 @@ def fetch_toi():
                 if n.wof_id in superseded_by_wof_ids:
                     diffs.append((n, None))
```
zerebubuth: Comparing this to the "funky" logic below, should be also be adding `n.wof_id` to `ids_to_delete` when it's superseded?

rmarianski: Yea, but that happens above in the [failure.superseded check](https://github.com/mapzen/tilequeue/pull/51/files#diff-3c589dc895a7fd4ecd09a8c77c9cb336L866). The superseded case is a little pedantic because it _should_ have been caught earlier because that's in the meta csv file. This scenario should only trigger when it's not listed as superseded in the meta file but it is in the json record, so we know it needs to be a removal, or otherwise we wouldn't have fetched the raw record. It should be equivalent though to just move that remove line down to make it look more similar to the funky case. I'll do that to make it look clearer.

Adding a little more detail, the funky case is a little bit different because that's not tracked in the meta csv file. So if the raw json is marked as funky it can either mean that an existing record became funky, or that we don't have that record and we should just not add it. That's why I had to guard adding it to the removal list a little bit more. I suppose it's fine to always issue deletes for all funky records even if they don't exist in our database, but it seemed wasteful to do that on every run.

rmarianski: Updated in: https://github.com/mapzen/tilequeue/commit/0757aeaf803530e9c175daddceae0fa3c5733469

zerebubuth:
:+1:

PR #56, opened 2015-11-25
Update s3 path

rmarianski:
Refs https://github.com/mapzen/tilequeue/issues/31
Connects with https://github.com/mapzen/tileserver/pull/10

Prefix the s3 path with a date and the first 5 characters of the url md5 hash

### DIFF
--- a/tilequeue/store.py
+++ b/tilequeue/store.py
@@ -2,20 +2,54 @@
 
 from boto import connect_s3
 from boto.s3.bucket import Bucket
-from TileStache.S3 import tile_key
+import md5
 import os
+
+
+def calc_hash(s):
+    m = md5.new()
+    m.update(s)
+    md5_hash = m.hexdigest()
+    return md5_hash[:5]
+
+
+def s3_tile_key(date, path, layer, coord, extension):
+    path_to_hash = '/%(path)s/%(layer)s/%(z)d/%(x)d/%(y)d.%(ext)s' % dict(
+        path=path,
+        layer=layer,
+        z=coord.zoom,
+        x=coord.column,
+        y=coord.row,
+        ext=extension,
+    )
+    md5_hash = calc_hash(path_to_hash)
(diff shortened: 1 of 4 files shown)


tilequeue/store.py
```diff
@@ -2,20 +2,54 @@
...
+    path_to_hash = '/%(path)s/%(layer)s/%(z)d/%(x)d/%(y)d.%(ext)s' % dict(
+        path=path,
+        layer=layer,
+        z=coord.zoom,
+        x=coord.column,
+        y=coord.row,
+        ext=extension,
+    )
+    md5_hash = calc_hash(path_to_hash)
+    s3_path = ('%(date)s/%(md5)s/'
```
zerebubuth: Can we re-use `path_to_hash` here? I.e:

```python
s3_path = ('%(date)s/%(md5)s/%(path_to_hash)s' % dict(
    date=date,
    md5=md5_hash,
    path_to_hash=path_to_hash)
```

Just in case we change how we want to hash the path in the future, we don't have to worry about the two getting out of sync.

tilequeue/store.py
```diff
@@ -2,20 +2,54 @@
...
+                   md5=md5_hash,
+               ))
+    return s3_path
+
+
 class S3(object):
 
-    def __init__(self, bucket, layer_name, path='', reduced_redundancy=False):
+    def __init__(
+            self, bucket, layer_name, date_prefix, path, reduced_redundancy):
```
zerebubuth: Do we not need the defaults on `path` and `reduced_redundancy` any more?

zerebubuth: Oh, wait, yes. I see. I should have read all the way to the bottom before commenting.

tilequeue/store.py
rmarianski: Good call, updated in 2a8d93e

zerebubuth:
:+1:

PR #57, opened 2015-12-23
Fall back to using the geometry position

zerebubuth:
When the label position is missing. This isn't ideal, but the geometry position will suffice until the record is fixed.

The previous code would error if the label position was missing and, while missing the label position isn't great, it can be recovered from. The error, on the other hand, causes alerts, which is a bit extreme.

@nvkelso could you review, please?


nvkelso:
LGTM

PR #64, opened 2016-01-13
Add support for clipping against configurably-padded bounds.

zerebubuth:
Previously, the choice was either to clip each feature against the bounds of the tile, or not to clip it. This adds a new parameter, `clip_factor`, which allows the size of the clipping box to be scaled. The scaling means that, for example, a `clip_factor` of 1.5 would mean that the clipping box was 50% larger than the original box (25% to each side). The expansion keeps the centroid the same.

Connects to mapzen/vector-datasource#197.

@nvkelso could you review, please?


nvkelso:
Do we need this one any longer if we're clipping purely to tile bounds for buildings? Is there any other reason to keep it around?

bcamper:
I'm not sure we're ready to make that call yet?

zerebubuth:
The code is written, so we may as well merge it even if we're not using it. It's possible, as @bcamper says, that we might revisit this for buildings, or find that it's useful for some other feature that we don't yet know about. The only downside I can see to merging it is if it introduces a maintenance overhead, but it's really not that much code and didn't involve changing any interfaces.

nvkelso:
Code looks good to me, modulo buffering 3 (shows 1 now).

PR #65, opened 2016-01-25
Correct the s3 tile key path

rmarianski:
This eliminates the extra slash from the s3 path. I don't think that this would make any difference, because it seems like the objects were still being saved in the correct location regardless. But this change probably makes sense to make anyway.


zerebubuth:
:+1: looks good to me!

rmarianski:
I added a / prefix to the key name. It shouldn't affect anything, and the other store implementations don't call the `s3_tile_key` function. @zerebubuth still good?

zerebubuth:
Yup, still looks good.

PR #66, opened 2016-02-01
Update make_queue signature

rmarianski:
Updates the signature to be more flexible to support usage from
tileserver.


zerebubuth:
:+1: looks good to me.

PR #69, opened 2016-02-19
Pass bounds to post-process functions.

zerebubuth:
Pass unpadded bounds to any post-process function with 'bounds' as a kwarg.

This seemed preferable to passing it to all functions and having to change the function signature of all the functions. Although if we find ourselves doing this again, might be better to pass a context object which we can extend.

Connects to mapzen/vector-datasource#502.

### DIFF
--- a/tilequeue/process.py
+++ b/tilequeue/process.py
@@ -11,6 +11,7 @@ from tilequeue.transform import calculate_padded_bounds
 from TileStache.Config import loadClassPath
 from TileStache.Goodies.VecTiles.server import make_transform_fn
 from TileStache.Goodies.VecTiles.server import resolve_transform_fns
+from inspect import getargspec
 
 
 def _preprocess_data(feature_layers, shape_padded_bounds):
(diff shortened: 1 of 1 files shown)


tilequeue/process.py
```diff
@@ -65,12 +66,18 @@ def _preprocess_data(feature_layers, shape_padded_bounds):
...
     for step in post_process_data:
         fn = loadClassPath(step['fn_name'])
         params = step['params']
 
+        # inspect the function to tell if it has kwarg called 'bounds'
+        # and, if so, update the params to include the tile coordinate.
+        if 'bounds' in getargspec(fn).args:
+            params = params.copy()
+            params['bounds'] = bounds
+
```
rmarianski: We might want to explicitly just pass bounds to every single function, or grow a context/state/env variable that has got it in it and always pass everything along there.

rmarianski: Ah nm, saw that you had the same exact thought in the pr description :)

rmarianski:
:+1:

tilequeue/process.py
zerebubuth: Yeah, the question is, do we want to do that _now_ or wait until later... I've added #70 to track it.

PR #71, opened 2016-02-22
Allow transforms list to be optional

zerebubuth:
Use an empty list if the 'transforms' parameter is missing, rather than raise KeyError.

Connects to mapzen/vector-datasource#537.


rmarianski:
:+1:

Issue #72, opened 2016-02-22
Start meta-tiling at z16


rmarianski:
Refs https://github.com/mapzen/vector-datasource/issues/478

PR #73, opened 2016-02-22
Metatile at z16

rmarianski:
Connects to #72


zerebubuth:
Should we change the expiry zooms in `wof.py` as well?

rmarianski:
Nice catch! I think I got everything with 4fea62b.

zerebubuth:
:+1: looks good to me!

PR #74, opened 2016-02-26
Add road sort key tests

rmarianski:
Connects to https://github.com/mapzen/vector-datasource/issues/546

@nvkelso could you review please?

### DIFF
--- a/tests/test_transform.py
+++ b/tests/test_transform.py
@@ -20,29 +20,17 @@ class RoadSortKeyTest(unittest.TestCase):
         val = float(val)
         return self._call_fut({key: val}, zoom, shape, fid)
 
-    def test_tunnel(self):
-        sort_key = self._call_fut_with_prop('tunnel=yes')
-        self.assertEqual(295, sort_key)
-
     def test_layer_low(self):
         sort_key = self._call_fut_with_float_prop('layer=-5')
         self.assertEqual(300, sort_key)
 
+    def test_tunnel(self):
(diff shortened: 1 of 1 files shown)


rmarianski:
I pointed to a different tilestache branch in circleci temporarily. I'll remove that before merging the pr.

nvkelso:
Should these tests stay in this repo / special hard coded branch? 

How do we run them (is that stated in a README somewhere)?

nvkelso:
Otherwise looks good.

rmarianski:
I'd rather have the tests stay where they are, and pull the transform and sort functions out of the tilestache branch and into tilequeue. Or, we might want to just move them all to vector-datasource, make that into a python package, and have these tests run from there. Imho, in the longer run it makes more sense to have them in vector-datasource because they'll be in the same repo as the queries and sql functions, and they tend to change in concert.

Btw, I'll kill the custom branch from the circle config before merging. That's just a shim to get this pr to run the tests against the appropriate tilestache branch in the interim.

And tilequeue tests are run in the usual python way, `python setup.py test`. That's what circle currently does. The tilequeue repo itself doesn't have any docs, but this repo itself is very Mapzen specific about how we process tiles offline, and geared towards our architecture. There are some more generally useful bits in there, which we could extract into its own python tile library, but we ourselves don't have much to gain from doing that at the moment.

nvkelso:
Sounds good with me.

Longer term I agree with moving more of this stuff (transforms, tests) into `vector-datasource` directly. But that's separate that this current work.

PR #75, opened 2016-02-29
Support storing individual formatted layers

rmarianski:
Connect to https://github.com/mapzen/tile-tasks/issues/43.

### DIFF
--- a/tilequeue/config.py
+++ b/tilequeue/config.py
@@ -169,8 +169,8 @@ def default_yml_config():
             'query-config': None,
             'template-path': None,
             'reload-templates': False,
-            'formats': ('json',),
-            'layers-to-format': (),
+            'formats': ['json'],
+            'layers-to-format': [],
         },
         'logging': {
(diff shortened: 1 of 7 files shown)


tilequeue/config.py
```diff
@@ -157,6 +170,7 @@ def default_yml_config():
             'template-path': None,
             'reload-templates': False,
             'formats': ('json',),
+            'layers-to-format': (),
```
zerebubuth: Should this be an empty tuple, or an empty list? Or doesn't it matter?

rmarianski: Doesn't matter here, because we only just iterate through them. It would if we tried to mutate those values anywhere though; we'd get a failure then.

But using lists might be more consistent though since that's how they would look when configured in yaml. I think I just saw the tuple on the line above and followed suit :) I made the updates in 1bab809.

zerebubuth:
:+1: looks great

PR #77, opened 2016-03-01
Pass context object to post-process functions.

zerebubuth:
Connects to mapzen/tilequeue#70.

### DIFF
--- a/tilequeue/command.py
+++ b/tilequeue/command.py
@@ -29,6 +29,7 @@ from tilequeue.worker import QueuePrint
 from tilequeue.worker import S3Storage
 from tilequeue.worker import SqsQueueReader
 from tilequeue.worker import SqsQueueWriter
+from TileStache.Config import loadClassPath
 from urllib2 import urlopen
 import argparse
 import logging
@@ -43,6 +44,7 @@ import time
 import yaml
 import datetime
 import os.path
+import traceback
 
 
 def create_command_parser(fn):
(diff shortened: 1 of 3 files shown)


zerebubuth:
@rmarianski had a much better idea than passing a cache around between functions.

I've implemented something based on that idea, where a post-process function config can have a `resources` section in addition to the `params` section. Each resource (for example, the one for the [csv matcher](https://github.com/mapzen/vector-datasource/blob/fe4f36e9eb57f1ed1b771b75b0b5ee832fd0a109/queries.yaml#L160-L164)) will have `type` and `init_fn` parameters which are used to load a function and pass it appropriate parameters - in the case of `type: file` that's an open file handle to the given `path`.

This seems much cleaner, as the file source is no longer assumed to be the filesystem in TileStache - it only exports the code which knows how to interpret the I/O object.

rmarianski:
:zap: :+1: 

Looks like the code in TileStache is then going to grab the resource under the particular name, yes?

I was originally thinking that tilequeue would do even less, and just call the factory function with the provided config. And then the function itself would return back the actual process step that would end up getting used.
But, I just realized that as it stands now, the result of this needs to get pickled because the processing happens in a different process. I might look into changing how some of this works in the future by adding an initialization step in the processing nodes where this sort of thing could happen, but I don't think it's worth it at this point.

zerebubuth:
Yeah, the CSVMatcher is a pure Python object with no I/O, so it's fine... but if we had something more complex it would make sense to initialise on a per-thread or per-process basis.

PR #78, opened 2016-03-04
Fix missing small buildings at z16

zerebubuth:
Don't simplify / remove small features at zoom 16, as this is now the zoom which contains everything needed for cutting higher level zoom tiles.

Connects to mapzen/vector-datasource#557.


rmarianski:
:+1:

PR #83, opened 2016-04-11
Delegate quantization to mapbox-vector-tile

rmarianski:
Connects to https://github.com/mapzen/tilequeue/issues/82

Depends on https://github.com/mapzen/TileStache/pull/141


zerebubuth:
:+1:

PR #85, opened 2016-05-04
Update json encoding to handle invalid geometries

rmarianski:
Connects to https://github.com/mapzen/vector-datasource/issues/698

Instead of using a custom json encoder to truncate precision, use the python `round` function.

I also added in other optimizations:

* avoid unnecessary serialization/deserialization to/from wkb for json and mvt.
* for json, instead of serializing and deserializing each layer and then serializing them all again, just make the appropriate structure and serialize once

### DIFF
--- a/tilequeue/format/__init__.py
+++ b/tilequeue/format/__init__.py
@@ -10,12 +10,14 @@ from TileStache.Goodies.VecTiles.topojson import merge as topojson_merge
 
 class OutputFormat(object):
 
-    def __init__(self, name, extension, mimetype, format_fn, sort_key):
+    def __init__(self, name, extension, mimetype, format_fn, sort_key,
+                 supports_shapely_geometry):
         self.name = name
         self.extension = extension
         self.mimetype = mimetype
         self.format_fn = format_fn
         self.sort_key = sort_key
+        self.supports_shapely_geometry = supports_shapely_geometry
 
     def __repr__(self):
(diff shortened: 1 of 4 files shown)


tilequeue/format/__init__.py
```diff
@@ -38,27 +40,21 @@ def format_tile(self, tile_data_file, feature_layers, coord, bounds_merc,
...
-        # merge function
-        deserialized_features = loads(out.getvalue())
-        layers.append(deserialized_features)
-    json_merge(fp, names, layers, None, coord)
+    else:
+        layer_features = []
+        for feature_layer in feature_layers:
+            layer_name = feature_layer['name']
+            features = feature_layer['features']
+            layer_features.append((layer_name, features))
```
zerebubuth: Why use a `list` of `tuple` here, rather than a `dict`? Is there some situation in which we have multiple layers with the same name? (If so, that's going to cause issues when we do MVT 2.0, as that doesn't allow it.)

zerebubuth:
:+1:

tilequeue/format/__init__.py
rmarianski: Just that it was only used for iterating and not lookups, so a list of tuples was more lightweight. But you're right that by taking a dict we ensure that it'll be a unique list of keys that get passed in, so we wouldn't have the situation where duplicates overwrote earlier layers in this case. Updated in eff37b1.

PR #87, opened 2016-05-12
Support configurable area thresholds

rmarianski:
Connects to https://github.com/mapzen/vector-datasource/issues/202

Adds support for a configurable area threshold value in pixels. This is a per layer value that will be applied to all polygonal features. Any feature whose area is less than this value will be dropped from tiles.

### DIFF
--- a/tilequeue/command.py
+++ b/tilequeue/command.py
@@ -371,6 +371,7 @@ def parse_layer_data(query_cfg, template_path, reload_templates, cfg_path):
     for layer_name, layer_config in layers_config.items():
         template_name = layer_config['template']
         start_zoom = layer_config['start_zoom']
+        area_threshold = int(layer_config.get('area-inclusion-threshold', 1))
         if reload_templates:
             query_generator = DevJinjaQueryGenerator(
                 environment, template_name, start_zoom)
@@ -388,6 +389,7 @@ def parse_layer_data(query_cfg, template_path, reload_templates, cfg_path):
             simplify_before_intersect=layer_config.get(
                 'simplify_before_intersect', False),
             simplify_start=layer_config.get('simplify_start', 0),
+            area_threshold=area_threshold,
         )
         layer_data.append(layer_datum)
         if layer_name in all_layer_names:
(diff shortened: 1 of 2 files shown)


tilequeue/command.py
```diff
@@ -361,6 +361,11 @@ def parse_layer_data(query_cfg, template_path, reload_templates, cfg_path):
     layer_data = []
     all_layer_data = []
     post_process_data = []
+    area_threshold = query_cfg.get('area-inclusion-threshold')
+    try:
+        area_threshold = int(area_threshold)
+    except (ValueError, TypeError):
+        area_threshold = 1
```
zerebubuth: Do we want to hide the error here, or do something more like:

```python
area_threshold = int(query_cfg.get('area-inclusion-threshold', 1))
```

And allow a config error to stop the process?

zerebubuth:
Is the first step to have a globally-configurable value, rather than one per layer?

rmarianski:
Did we ultimately want a separate setting per layer? If so, I think we might as well just go right to that.

zerebubuth:
I think so... or, at least, that's what I _thought_ @nvkelso was asking for. Perhaps I got it wrong?

rmarianski:
The only downside really is having to specify it multiple times if we always want it to be the same, but we gain more flexibility. I made it per layer in 004ca68.

tilequeue/command.py
```diff
@@ -377,6 +377,11 @@ def parse_layer_data(query_cfg, template_path, reload_templates, cfg_path):
         else:
             template = environment.get_template(template_name)
             query_generator = JinjaQueryGenerator(template, start_zoom)
+            layer_config.get('area-inclusion-threshold')
+            if layer_config is None:
+                area_threshold = 1
+            else:
+                area_threshold = int(layer_config)
```
zerebubuth: Why not spell this `area_threshold = int(layer_config.get('area-inclusion-threshold', 1))`?

rmarianski: Good call. 8bad4d3

zerebubuth:
:+1:

nvkelso:
Yep, configurable by layer. `water` might have a different value than `landuse`, etc.

PR #88, opened 2016-05-13
Add l10n_name to wof processing

rmarianski:
Connects to https://github.com/mapzen/vector-datasource/issues/418
Depends on https://github.com/mapzen/vector-datasource/pull/816

### DIFF
--- a/tilequeue/wof.py
+++ b/tilequeue/wof.py
@@ -1,13 +1,16 @@
 from collections import namedtuple
 from contextlib import closing
 from cStringIO import StringIO
+from datetime import datetime
 from operator import attrgetter
+from psycopg2.extras import register_hstore
 from shapely import geos
 from tilequeue.tile import coord_marshall_int
 from tilequeue.tile import coord_unmarshall_int
 from tilequeue.tile import mercator_point_to_coord
 from tilequeue.tile import reproject_lnglat_to_mercator
 import csv
+import edtf
 import json
 import os.path
 import psycopg2
@@ -17,7 +20,6 @@ import shapely.geometry
 import shapely.ops
 import shapely.wkb
 import threading
-import edtf
 
 
 def generate_csv_lines(requests_result):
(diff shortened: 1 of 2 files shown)


tilequeue/wof.py
```diff
@@ -330,10 +332,25 @@ def failure(reason):
...
+    lang_suffix_size = len('_preferred') * -1
+    l10n_names = {}
+    for k, v in props.iteritems():
+        if not v:
+            continue
+        if not k.startswith('name:') or not k.endswith('_preferred'):
+            continue
+        if isinstance(v, list):
+            v = v[0]
+        lang = k[:lang_suffix_size]
```
zerebubuth: Doesn't this take the _first_ `lang_suffix_size` chars, rather than drop the last chars?

zerebubuth: Oh, wait, I missed the `* -1`... would this be easier to read if it were `k[:-lang_suffix_size]` and `lang_suffix_size` were positive?

rmarianski: Yea, we'll get something that looks like 3 letter code_ wof name type. For example: `name:eng_x`

Would we ever care about the name type? What should we store on our end, just `name:eng`?

As it stands now, I'm taking only the preferred names, and just the first one at that. I'm expecting that in the tile we'll only show a single value for a language code, to match how it works with the osm tags. Should we take a different approach to select the best candidate?

zerebubuth: Agreed - I'm not sure it's worth having more than one value for the names. I'd say let's do the simple thing now and if anyone finds themselves wanting `alt_name` then we can do that in a follow-up ticket.

rmarianski: Good point on making the suffix trimming clearer. Updated in 7b32389.

zerebubuth:
:+1:

nvkelso:
Agree on limiting to first "preferred" localized WOF name, and  trimming off the junk (resulting in `name:eng` not `name:eng_x`). 

FYI: Aaron and I are discussing some options in WOF for alt names and specifying name preference order (including official & spoken), but all that's for followup work probably in v1.1 or later.

rmarianski:
I added some tests for the transform itself in f8affa8.

PR #90, opened 2016-05-26
Port relevant tilestache fork code into tilequeue

rmarianski:
Connects to https://github.com/mapzen/vector-datasource/issues/211


zerebubuth:
Code looks okay to me, but were you expecting the test failure?

tilequeue/process.py
```diff
@@ -5,15 +5,30 @@
 from tilequeue.tile import coord_to_mercator_bounds
 from tilequeue.tile import pad_bounds_for_zoom
 from tilequeue.tile import tolerance_for_zoom
-from tilequeue.transform import mercator_point_to_wgs84
+from tilequeue.transform import mercator_point_to_lnglat
 from tilequeue.transform import transform_feature_layers_shape
 from tilequeue.transform import calculate_padded_bounds
 from TileStache.Config import loadClassPath
```
zerebubuth: If we've removed TileStache from the dependencies in `setup.py`, do we also need to drop this import and have a local copy of `loadClassPath`?

rmarianski: Ah, good catch. I must have still had it in my path which is why it continued to work for me.

Is it cool to replace that with a different library that only supported loading from a dotted path? One that I've used in the past is [zope.dottedname](https://pypi.python.org/pypi/zope.dottedname).

rmarianski: I went ahead and made that change in 51bd120.

rmarianski:
The test failure was unexpected. I removed the [tests/__init__.py file](b6ca0cb), and now it works on circle. Odd that I didn't see that locally though.

zerebubuth:
:+1:

Great work!

PR #92, opened 2016-06-14
Add configurable buffer support

rmarianski:

The bounding box that the queries used is buffered to be the max of any of the buffer sizes configured across all formats. This trade-off allowed us to still use a single bounding box in the queries to keep things simpler for now. We could be more precise here, but we would need to update all queries to use the corresponding bounding box to match the geometry type.

Another potential performance issue is that the buffered bounds are calculated from the config per feature now, because the geometry type is a factor. This could be made faster if need be.

There are also some tileserver updates that are required. I'll put in a pr for that once the tilequeue changes are nailed down. There will also be updates necessary to the chef recipes that generate the config files.

Unfortunately, I also introduced some (more) circular imports, as well as importing some (more) functions that start with _ in other locations. I'd rather tackle this all at once separately since it was a problem before too.

### DIFF
--- a/tilequeue/query.py
+++ b/tilequeue/query.py
@@ -67,10 +67,7 @@ def build_feature_queries(unpadded_bounds, layer_data, zoom):
     queries_to_execute = []
     for layer_datum in layer_data:
         query_bounds_pad_fn = layer_datum['query_bounds_pad_fn']
-        if query_bounds_pad_fn:
-            padded_bounds = query_bounds_pad_fn(unpadded_bounds)
-        else:
-            padded_bounds = unpadded_bounds
+        padded_bounds = query_bounds_pad_fn(unpadded_bounds)
         query_generator = layer_datum['query_generator']
         query = query_generator(padded_bounds, zoom)
         queries_to_execute.append((layer_datum, query, padded_bounds))
--- a/tilequeue/query.py
+++ b/tilequeue/query.py
@@ -1,7 +1,6 @@
 from psycopg2.extras import RealDictCursor
 from tilequeue.postgresql import DBAffinityConnectionsNoLimit
 from tilequeue.tile import coord_to_mercator_bounds
-from tilequeue.tile import pad_bounds_for_zoom
 import sys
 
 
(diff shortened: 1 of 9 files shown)


tilequeue/query.py
```diff
@@ -64,21 +63,26 @@ def jinja_filter_bbox(bounds, srid=900913):
...
 
-def build_feature_queries(bounds, layer_data, zoom):
+def build_feature_queries(unpadded_bounds, layer_data, zoom):
     queries_to_execute = []
     for layer_datum in layer_data:
+        query_bounds_pad_fn = layer_datum['query_bounds_pad_fn']
+        if query_bounds_pad_fn:
+            padded_bounds = query_bounds_pad_fn(unpadded_bounds)
+        else:
+            padded_bounds = unpadded_bounds
```
zerebubuth: `_create_query_bounds_pad_fn` returns `_bounds_pad_no_buf` when not configured, which is passed directly into the `dict` when making the layer datum. Can `query_bounds_pad_fn` ever be false-y?

tilequeue/query.py
```diff
@@ -64,21 +63,26 @@ def jinja_filter_bbox(bounds, srid=900913):
     return bbox
 
 
-def build_feature_queries(bounds, layer_data, zoom):
+def build_feature_queries(unpadded_bounds, layer_data, zoom):
     queries_to_execute = []
     for layer_datum in layer_data:
+        query_bounds_pad_fn = layer_datum['query_bounds_pad_fn']
+        if query_bounds_pad_fn:
+            padded_bounds = query_bounds_pad_fn(unpadded_bounds)
```
zerebubuth: Are the `unpadded_bounds` are in mercator coords? Should we be applying some sort of conversion factor to in the padding process, since the config file has them in 1/256ths of a tile?

rmarianski: Doh. Nice catch! The configured numbers should be in pixels, and I've been converting it to meters in other areas by multiplying by the meters per pixel for the given zoom level. I'll go ahead and update.

tilequeue/query.py
rmarianski: Don't think so, as in I think functions are always truthy. I could alternatively make it return none to represent no buffer operation, but this seemed easier at the time. Is that what you had in mind?

zerebubuth: If the `else` branch can't be taken, and the default `query_bounds_pad_fn` is doing a no-op anyway, could we lose the `if` altogether and simplify the code?

rmarianski: Ah, of course. I missed that you were saying this earlier. I'll go ahead and make this change too.

rmarianski: Updated in b01bef9.

tilequeue/query.py
rmarianski: Updated in fc56465.

rmarianski:
I also passed the padded bounds through the processing pipeline in a00eec7. This gives the post processing transforms access to it.

rmarianski:
Fixed the tests in 964d421.

zerebubuth:
:+1:

PR #94, opened 2016-06-15
Distinguish between meters per pixel area vs dim

rmarianski:


zerebubuth:
:+1: looks great!

PR #96, opened 2016-06-17
Ensure json properties are utf8 encoded

rmarianski:
For consistency, keep json property strings encoded in utf8.

### DIFF
--- a/tilequeue/process.py
+++ b/tilequeue/process.py
@@ -69,7 +69,8 @@ def _preprocess_data(feature_layers):
                         if output_val is not None:
                             # all other tags are utf8 encoded, encode
                             # these the same way to be consistent
-                            output_key = output_key.encode('utf-8')
+                            if isinstance(output_key, unicode):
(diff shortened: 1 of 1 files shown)


rmarianski:
The strings from json properties were coming back through as unicode. This pr encodes them to utf8 to be consistent with the rest of the tags.

Alternatively, we could ensure all strings are unicode as soon as possible, which probably makes more sense in the long run, but this way is less invasive and less risky in the short term.

zerebubuth:
Is `output_key` always a `unicode`? I'm just wondering why the `isinstance()` check for the value, but not the key?

rmarianski:
> Is output_key always a unicode? I'm just wondering why the isinstance() check for the value, but not the key?

Yes, the key always is. I can put in a check there though for clarity purposes.
For the values, we can get numeric values back sometimes which we don't want to encode, so that needs to be there.

rmarianski:
Updated in 4711372.

zerebubuth:
Looks great! :+1:

PR #97, opened 2016-06-21
Discriminate query bounds on geometry type

rmarianski:
Depends on https://github.com/mapzen/vector-datasource/pull/870

The query or `padded_bounds` are now a dict with 3 bounds keyed off of
geometry type.


zerebubuth:
Circle says the imports in the test need fixing, otherwise it's :+1: from me.

rmarianski:
Updated tests in ff66174.

PR #102, opened 2016-08-01
Update wof projection to 3857

rmarianski:
Connects to https://github.com/tilezen/vector-datasource/issues/908

This should only get merged in once the hstore-all branch in vector-datasource does.

### DIFF
--- a/tilequeue/query.py
+++ b/tilequeue/query.py
@@ -38,7 +38,7 @@ def jinja_filter_geometry(value):
     return 'ST_AsBinary(%s)' % value
 
 
-def jinja_filter_bbox_filter(bounds, geometry_col_name, srid=900913):
+def jinja_filter_bbox_filter(bounds, geometry_col_name, srid=3857):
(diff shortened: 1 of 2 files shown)


zerebubuth:
Is it worth pulling 3857 out as `DATABASE_SRID`, a module level constant or something like that?

rmarianski:
Sure, is ffcb329 what you had in mind?

rmarianski:
I snuck in a fix for data insertion in cbc5166.

zerebubuth:
Looks great, :+1:

zerebubuth:
Oh, nice catch! Do we need that anywhere else as well? Presumably all the OSM tables are okay, since `osm2pgsql` must be doing something similar?

rmarianski:
Turns out that 45099b6 is also needed.

I think this is specific to the initial wof load and the `copy` format. I expect other places should be ok if they go through the driver.

PR #103, opened 2016-08-05
Add basic size logging for the objects returned from queries.

zerebubuth:
This will allow us to collect statistics about which layers contribute most to network traffic, and may help us optimise queries which are returning too much data.

### DIFF
--- a/tilequeue/query.py
+++ b/tilequeue/query.py
@@ -3,7 +3,6 @@ from tilequeue.postgresql import DBAffinityConnectionsNoLimit
 from tilequeue.tile import calc_meters_per_pixel_dim
 from tilequeue.tile import coord_to_mercator_bounds
 import sys
-import logging
 
 
 def generate_query(start_zoom, template, bounds, zoom):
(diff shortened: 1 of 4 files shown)


tilequeue/query.py
```diff
@@ -82,6 +83,14 @@ def execute_query(conn, query, layer_datum, padded_bounds):
...
         cursor.execute(query)
         rows = list(cursor.fetchall())
+
+        # if we are debugging this layer, then print out the size of the
+        # results returned from the server.
+        if layer_datum.get('debug_query_size'):
+            logger = logging.getLogger('query')
+            logger.debug('Layer[%r]: %d bytes' %
+                         (layer_datum.get('name'), sys.getsizeof(rows)))
+
```
rmarianski: The one gotcha is that I think the way the geometry gets encoded is into a reader type of object. The main thread that aggregates the results then explicitly [calls bytes()](https://github.com/tilezen/tilequeue/blob/0aed62cea12038230768f914569c1c7c9c677594/tilequeue/query.py#L175-L186) on it to convert it to bytes so that it can be pickled when passing it to the processors.

In any event, it's possible that this will report the correct number of bytes, but I'm not sure.

rmarianski:
> Would it be better to count this at a higher level (e.g: once we have Features) or perhaps at a lower level (e.g: as @rmarianski suggested by pickling the objects and counting the size of the resulting buffer)?

Good catch on it being a shallow size. Maybe we just pickle the rows themselves after they have all been [read in](https://github.com/tilezen/tilequeue/blob/0aed62cea12038230768f914569c1c7c9c677594/tilequeue/query.py#L186) and print the size out there?

zerebubuth:
I moved the counting logic to the `_preprocess_data` function and strung the results through the `metadata` item which is currently used for timing information. Will that break logfile parsing if it's output in the regular worker output?

Also: requires a corresponding change in tileserver, if we think this is the right way to go.

rmarianski:
:+1:

I double checked the log parsing and I think we should be ok because it's at the end of the line. We can verify once deployed on dev that we're continuing to receive data.

PR #104, opened 2016-08-07
Extract size metadata from layers key

rmarianski:


zerebubuth:
Oops! Thanks for catching this. :+1:

PR #105, opened 2016-08-10
move VERSION to file, look that up in setup.py

nvkelso:
In the future we would have called this `0.11.0-pre1` instead of `0.11.0.dev0` to follow SemVer, but since dev0 has been released, we don't touch that.

@rmarianski looks like we need `0.11.0.dev0` notes in the changelog? Can you do that in this same PR and then when it's merged we can tag it and add formal release notes via the Github UI?


rmarianski:
0.11.0.dev0 was never released or tagged, we've just been updating the setup.py version after releases on this branch to signal that the next version is in development. This is useful so the version specifier on the tagged release only exists in one commit. That specifier was chosen because that's what [python honors](https://www.python.org/dev/peps/pep-0440/#post-release-separators) and has meaning for dependency resolution.

In practice it might not matter because we ourselves never actually install from python packages, but it would for mapbox-vector-tile because we do push releases to [pypi](https://pypi.python.org/pypi/mapbox-vector-tile).

PR #106, opened 2016-08-11
Support multiple geometries in queries

rmarianski:
Connects to https://github.com/tilezen/vector-datasource/issues/965

@zerebubuth could you review please?

PR #107, opened 2016-08-12
Support multiple geometries in queries

rmarianski:
Refs https://github.com/tilezen/tilequeue/pull/106

This pulls in the changes from multiple-geometries-queries rebased onto master.
This should also demonstrate that there aren't additional test failures.


zerebubuth:
Yup, cool. That's looking good :+1:

Issue #80, opened 2016-04-04
Don't re-PUT unchanged items


zerebubuth:
At the moment, a tile is always stored in S3 after it is rendered. The majority of renders over the long term are caused by tile expiries. `osm2pgsql` has to be very conservative when generating tile expiries, and will expire features on which any tag has changed or the whole interior and/or boundary of large features if any point on them changes. This can potentially lead to re-rendering tiles with the exact same result as the existing tile.

Filtering these out before rendering seems pretty hard - it would mean examining the changes to see if they match the queries and survive the tile transform pipeline.

However, filtering them out just before storing them would be pretty easy: Do a GET request just before the PUT and if the content is identical then don't do the PUT. In the simple case this is a win if 8% or more of re-renders are identical, but that percentage gets lower the more we amplify our writes with more formats and replica regions.

bcamper:
Nice idea.

nvkelso:
This is live on prod now.

PR #108, opened 2016-08-18
Conditionally store tile data

rmarianski:
Connects to https://github.com/tilezen/tilequeue/issues/80

Only store the tile data if it's different from the existing data in the
store.


zerebubuth:
Nice! :+1:

Here's hoping it makes a big difference.

rmarianski:
@zerebubuth I put in updated logic for the counting, which will log total stored and not when processing a tile coordinate.

I had originally thought this would be useful for just child coordinates, but each tile coordinate stores per format as well which I overlooked previously, so we should get more accurate counts here. We might want to consider not checking the rest of the formats if the first succeeds, but always checking is safer in the case where some formats succeed and others don't. If we bailed early we could end up in a situation where the data for some formats would be stale.

zerebubuth:
Looks great :+1:

PR #111, opened 2016-08-25
Correct the buffered mvt format

rmarianski:
Connects to https://github.com/tilezen/vector-datasource/issues/106

The previous code seemed like it would override the mvt format that got stored with the buffered format because the extension was the same. That seems odd because I thought we had checked the .mvtb extension, but maybe not?

@nvkelso, @zerebubuth could you review please?


zerebubuth:
:+1:

PR #120, opened 2016-11-15
Add a command to dump the tiles of interest list to a text file

iandees:
For https://github.com/mapzen/tile-tasks/issues/118

### DIFF
--- a/tilequeue/command.py
+++ b/tilequeue/command.py
@@ -1056,8 +1056,6 @@ def tilequeue_dump_tiles_of_interest(cfg, peripherals):
     n_toi = len(coords)
     logger.info('Fetching tiles of interest ... done')
 
-    from tilequeue.tile import serialize_coord
-
     toi_filename = "toi.txt"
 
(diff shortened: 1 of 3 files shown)


tilequeue/command.py
```diff
@@ -1047,6 +1047,28 @@ def tilequeue_initial_load_wof_neighbourhoods(cfg, peripherals):
...
+    logger.info('Fetching tiles of interest ...')
+    coords = peripherals.redis_cache_index.fetch_tiles_of_interest()
+    n_toi = len(coords)
+    logger.info('Fetching tiles of interest ... done')
+
+    toi_filename = "toi.txt"
+
+    logger.info('Writing %d tiles of interest to %s ...', n_toi, toi_filename)
+    with open(toi_filename, "w") as f:
+        for coord in map(coord_unmarshall_int, coords):
```
rmarianski: It would be better to use either `imap` here or call the function during the iteration.

PR #121, opened 2016-11-15
Don't use map during forloop while dumping TOI

iandees:
For https://github.com/tilezen/tilequeue/pull/120#discussion_r88032484


iandees:
@rmarianski good catch in https://github.com/tilezen/tilequeue/pull/120#discussion_r88032484. How's this?

PR #122, opened 2016-11-16
Add a command to load tiles of interest from file

iandees:
To help with https://github.com/mapzen/tile-tasks/issues/118.
The reverse of #120.


tilequeue/command.py
```diff
@@ -1070,6 +1070,50 @@ def tilequeue_dump_tiles_of_interest(cfg, peripherals):
...
+        app.logger.info("Chunk of {} done".format(chunk_size))
+
+    logger.info(
+        'Adding tiles of interest from %s to key %s... done',
+        toi_filename,
+        new_toi_key,
+    )
+
+    # Rename the new TOI key to the old TOI key
+    # redis_client.rename(new_toi_key, cfg.redis_cache_set_key)
```
iandees: Note that I'm still testing this.

tilequeue/command.py
```diff
@@ -1070,6 +1070,50 @@ def tilequeue_dump_tiles_of_interest(cfg, peripherals):
...
+def tilequeue_load_tiles_of_interest(cfg, peripherals):
+    """
+    Given a newline-delimited file containing tile coordinates in
+    `zoom/column/row` format, load those tiles into a new Redis SET
+    alongside the existing one, then RENAME the new to the old.
+    """
+    logger = make_logger(cfg, 'load_tiles_of_interest')
+    logger.info('Loading tiles of interest')
+
+    toi_filename = "/Users/iandees/Downloads/toi.txt"
```
iandees: I'd really like to have a way to pass a command line arg into the command without having to add it to the config file. Do we support something like that?

rmarianski: Yes, ironically we started off that way, but we started moving everything to the config yaml because the command lines got unwieldy and we were repeating a lot most of the time.

Instead of calling `create_command_parser`, I'd suggest making your own version. There's also a little bit in the configuration that tries to merge what would have been passed on the command line with the yaml, but that hasn't been exercised in a bit.

tilequeue/command.py
```diff
@@ -1070,6 +1070,84 @@ def tilequeue_dump_tiles_of_interest(cfg, peripherals):
...
+def tilequeue_load_tiles_of_interest(cfg, peripherals):
+    """
+    Given a newline-delimited file containing tile coordinates in
+    `zoom/column/row` format, load those tiles into a new Redis SET
+    alongside the existing one, then RENAME the new to the old.
+    """
+    logger = make_logger(cfg, 'load_tiles_of_interest')
+    logger.info('Loading tiles of interest')
+
+    toi_filename = "/Users/iandees/Downloads/tile_requests_unique_noeurope.txt"
```
rmarianski: hard coded path

tilequeue/command.py
```diff
@@ -1070,6 +1070,84 @@ def tilequeue_dump_tiles_of_interest(cfg, peripherals):
...
+    chunk_size = 10000
+    new_toi_key = cfg.redis_cache_set_key + "-new"
+    old_toi_key = cfg.redis_cache_set_key + "-old"
+
+    def grouper(iterable, n, fillvalue=None):
+        """Collect data into fixed-length chunks or blocks"""
+        args = [iter(iterable)] * n
+        return izip_longest(fillvalue=fillvalue, *args)
```
rmarianski: I'm surprised this worked, because it's a non keyword arg after a keyword arg, but I guess the `*args` is special and python is ok with it. I would have expected to have to write: `izip_longest(*args, fillvalue=fillvalue)`.
(Nothing to change here, just thinking out loud) :)

tilequeue/command.py
nvkelso: @iandees Can you make the filename configurable, please?

iandees: I'm not sure how to fix this. See [this thread above](https://github.com/tilezen/tilequeue/pull/122#discussion_r88307679), but I don't really want to add a thing to the config file. I'd like to accept the filename at the command line.

nvkelso: See related discussion in: https://github.com/tilezen/tilequeue/pull/122#discussion_r88307679

tilequeue/command.py
zerebubuth: I notice that the code to dump TOI uses an output filename of `"toi.txt"` in the current working directory. While I would also like to be able to pass the filename on the command line, would a pragmatic solution for now be to use the same name?

zerebubuth:
I removed the local path in 221f665, loading instead from `"toi.txt"`. @rmarianski, @iandees - please could you look over the code and let me know what you think?

iandees:
I'm going to merge this because Matt fixed the hardcoded filename problem (for now).

PR #123, opened 2016-11-25
Add support for writing metatiles

zerebubuth:
If configured (`metatile size` is not `None`), this will write metatiles to storage rather than individual tiles. Each metatile is a zip file containing multiple formats, compatible with [Tapalcatl](https://github.com/tilezen/tapalcatl). Space for future extensions to multiple coordinates has been left, but isn't supported in this change.

I couldn't find anywhere that support for _reading_ metatiles is required in tilequeue at the moment. Is that only needed in tileserver?

@rmarianski, @iandees could you review, please?

### DIFF
--- a/tilequeue/metatile.py
+++ b/tilequeue/metatile.py
@@ -1,0 +1,31 @@
+import zipfile
+import StringIO
+from tilequeue.format import zip_format
+
+
+def make_metatiles(size, tiles):
+    """
+    Make a list of metatiles from a list of tiles.
+    """
+
+    assert size == 1, \
+        "Tilequeue only supports metatiles of size one at the moment."
+
+    if len(tiles) == 0:
+        return []
+
+    coord = tiles[0]['coord']
+    layer = tiles[0]['layer']
+
+    buf = StringIO.StringIO()
+    with zipfile.ZipFile(buf, mode='w') as z:
+        for tile in tiles:
+            assert tile['coord'] == coord
+            assert tile['layer'] == layer
+
+            tile_name = '0/0/0.%s' % tile['format'].extension
+            tile_data = tile['tile']
+            z.writestr(tile_name, tile_data)
+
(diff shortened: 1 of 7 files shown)


nvkelso:
Do we need to add support to Scoville or will the new metatile server act as middleware?

zerebubuth:
This change should be completely transparent to Scoville and any other users. Eventually, I think we'd want to make changes to Scoville to exercise different paths through the code. But I think we should wait until after we're done shuffling the code around so that we know what those paths will be.

tilequeue/metatile.py
```diff
@@ -0,0 +1,51 @@
...
+    coord = tiles[0]['coord']
+    layer = tiles[0]['layer']
+
+    buf = StringIO.StringIO()
+    with zipfile.ZipFile(buf, mode='w') as z:
+        for tile in tiles:
+            assert tile['coord'] == coord
+            assert tile['layer'] == layer
+
+            tile_name = '0/0/0.%s' % tile['format'].extension
```
rmarianski: Are we using a relative addressing coordinate scheme? (Just making sure that's the case here)

zerebubuth: Yup. I thought about whether we'd want it absolute or relative, and the only real difference is if we want "relocatable" metatiles. This might be useful if, for example, we have an "sea" one containing only an ocean rectangle. I couldn't see any downsides to making it relative, but I could easily be overlooking something. What do you think?

I hope we can support metatiling in `x`, `y` and possibly even `z` directions in a future version. But since none of these are supported in this version, I figured it was safe to hard-code this to `0/0/0`. :wink:

rmarianski: > I couldn't see any downsides to making it relative, but I could easily be overlooking something. What do you think?

Other than the obvious that consumers need to translate between the two, I can't think of any downsides either. And like you mentioned, relative coordinates could allow for more flexibility. :+1:

Issue #18, opened 2015-07-31
Explore adding priorities to tile processing


rmarianski:
All tiles are currently processed without any priority given to them. Priorities would allow us to control processing in more flexible ways, and may even increase throughput.

One way to add this would be to add support for multiple queues, where the queues are queried in a configured order. We would need to update:
- intersect process to enqueue tiles to the appropriate queue
- worker processor that pulls work from the queue should ask for tiles from queues in priority order
- seed process - probably best to seed tiles using the same strategy

Assuming we always want to have tiles enqueued using this strategy, it would be best to create a new tilequeue queue implementation that had this logic baked in. The other components would not need to change then. Even if we did need to support different behavior based on context, we could add in keyword arguments that certain queue implementations could honor, and others would ignore.

nvkelso:
Straw man proposal:

- Escalation queue immediately
- Zooms 15-16+ ASAP after escalation queue
- Zooms 13-14 hourly
- Zooms 11-12 daily
- Zooms 0-10 weekly

nvkelso:
Add pager duty alert for items in queue for longer than 1 week.

pnorman:
> Straw man proposal:

A generic ability to have queues sounds more useful - some deployments could have different tiers of customers

nvkelso:
@pnorman definitely – the priority queues should be initially for normalizing data updates but should be extendable towards different "areas of interest" and frequency cycles per customer interest. We could even implement tile pack support this way.

nvkelso:
- SQS pricing is $0.40 per million requests ($0.00000040 per request). 
- A single request can have from 1 to 10 messages, up to a maximum total payload of 256 KB.
- See more: https://aws.amazon.com/sqs/pricing/

Looks like we do 250k to 500k per day on average (but spike up to 6mm one day in last 2 weeks). **Let's say 1mm / day per queue = $12 / month.** 

If we have **5 queues** in the new priority system then that'd be **$60 / month**. Our prod deploy costs would remain consistent (and they'd just be in a single priority queue). 

We could adjust these costs by changing a time interval, but estimated cost is low enough to not worry about that now.

pnorman:
What about those not running tilequeue on AWS?

nvkelso:
@pnorman Mapzen uses AWS to run Tilequeue currently and will for the foreseeable future. How are you implementing this bit if you're not on AWS?

pnorman:
[...]

PR #124, opened 2016-12-01
Add priority queue implementation

rmarianski:
Connects to https://github.com/tilezen/tilequeue/issues/18

### DIFF
--- a/config.yaml.sample
+++ b/config.yaml.sample
@@ -1,7 +1,7 @@
 queue:
   # type can be `sqs`, `multisqs`, `file`, `mem`, or `redis`
   type: mem
-  name: <sqs queue/queue file name/redis key/list of queue names>
+  name: <sqs queue/list of queue names/queue file name/redis key/>
   # for multisqs
   zoom-queue-map:
     # keys are <start-zoom>-<end-zoom> (inclusive) ranges
--- a/config.yaml.sample
+++ b/config.yaml.sample
@@ -1,7 +1,13 @@
 queue:
-  # type can be `sqs`, `file`, `mem`, or `redis`
+  # type can be `sqs`, `multisqs`, `file`, `mem`, or `redis`
   type: mem
-  name: <sqs queue/queue file name/redis key>
+  name: <sqs queue/queue file name/redis key/list of queue names>
+  # for multisqs
+  zoom-queue-map:
+    # keys are <start-zoom>-<end-zoom> (inclusive) ranges
+    # values are the queue name to use for zooms in that range
+    0-10: queue-1
+    11-20: queue-2
 store:
   type: s3 # Can also be `directory`, which would dump the tiles to disk.
   name: <s3 bucket/tile directory name>
--- a/tilequeue/command.py
+++ b/tilequeue/command.py
@@ -99,11 +99,7 @@ class GetSqsQueueNameForZoom(object):
         return result
 
 
-def make_get_queue_name_for_zoom(queue_cfg, queue_names):
-    dispatch_cfg = queue_cfg.get('dispatch')
-    assert dispatch_cfg, 'Missing dispatch config for multiqueue'
-    assert isinstance(dispatch_cfg, dict), 'Dispatch queue cfg must be a map'
-
+def make_get_queue_name_for_zoom(dispatch_cfg, queue_names):
     zoom_to_queue_name_table = {}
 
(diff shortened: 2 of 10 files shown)


tilequeue/queue/multisqs.py
```diff
@@ -0,0 +1,181 @@
...
+            assert(left_to_read >= 0)
+            if left_to_read == 0:
+                break
+
+        return coord_messages
+
+    def job_done(self, coord_message):
+        queue_name = self.get_queue_name_for_zoom(coord_message.coord.zoom)
+        if queue_name is None:
+            # TODO log? should this assert instead?
```
zerebubuth: I think we should `assert` here. Presumably this would mean that tilequeue has rendered a zoom level for which it wasn't configured, which seems to me like an error condition. Should we assume that `get_queue_name_for_zoom` will return a valid queue name always, and have an assertion in that method (or use `[]` instead of `get()`)?

rmarianski: Added assert in 22ea268.

And upon further thought, I think that you're right that we'd want it to fail fast when it didn't find a queue for a particular input. That suggests either the mapping is wrong/incomplete, or the input coordinate is invalid, which I think we'd want to know about right away. I'll go ahead and make this change.

config.yaml.sample
```diff
@@ -1,7 +1,13 @@
 queue:
-  # type can be `sqs`, `file`, `mem`, or `redis`
+  # type can be `sqs`, `multisqs`, `file`, `mem`, or `redis`
   type: mem
-  name: <sqs queue/queue file name/redis key>
+  name: <sqs queue/queue file name/redis key/list of queue names>
```
nvkelso: This list and the list in the line above should be in the same order.

config.yaml.sample
```diff
@@ -1,7 +1,13 @@
...
+  # for multisqs
+  zoom-queue-map:
+    # keys are <start-zoom>-<end-zoom> (inclusive) ranges
+    # values are the queue name to use for zooms in that range
+    0-10: queue-1
```
nvkelso: Should the queue names be descriptive of the zooms, or generic?

tilequeue/command.py
```diff
@@ -87,9 +87,68 @@ def uniquify_generator(generator):
...
 
-def make_queue(queue_type, queue_name, redis_client,
+class GetSqsQueueNameForZoom(object):
+
+    def __init__(self, zoom_queue_table):
+        self.zoom_queue_table = zoom_queue_table
+
+    def __call__(self, zoom):
+        assert isinstance(zoom, int)
+        assert 0 <= zoom <= 20
```
nvkelso: We've standardized around `21` in house styles and Tangram.

tilequeue/command.py
```diff
@@ -87,9 +87,68 @@ def uniquify_generator(generator):
...
+        zoom_fields = zoom_range.split('-')
+        assert len(zoom_fields) == 2, 'Invalid zoom range: %s' % zoom_range
+        zoom_start_str, zoom_until_str = zoom_fields
+        try:
+            zoom_start = int(zoom_start_str)
+            zoom_until = int(zoom_until_str)
+        except (ValueError, KeyError):
+            assert not 'Invalid zoom range: %s' % zoom_range
+
+        assert (0 <= zoom_start <= 20 and
```
nvkelso: Ditto `21`

tilequeue/queue/sqs.py
```diff
@@ -68,9 +69,12 @@ def read(self, max_to_read=1):
             data = message.get_body()
             coord = deserialize_coord(data)
             if coord is None:
-                # log?
+                # TODO log?
```
nvkelso: ;)

config.yaml.sample
rmarianski: Good catch, updated in b6df7a3.

nvkelso:
(Discussed zoom 21 > 20 in Slack, we can keep as-is now. File new issue to
track that work as it’s more systemic / x-funcional in nature and doesn’t
directly relate to this PR.)

tilequeue/queue/multisqs.py
rmarianski: Updated in c4749ef.

rmarianski:
Tracking the zoom 20 -> 21 bump in https://github.com/mapzen/tile-tasks/issues/152

PR #132, opened 2016-12-14
Increase topojson scale from 1024 -> 4096

rmarianski:
Connects to https://github.com/tilezen/vector-datasource/issues/1145

Bumping the scale makes the rendering artifact go away. We'll likely want to revisit the topojson format again with an eye towards returning ogc valid geometries, but this is a quick fix that seems to    help. The downside here is that it may increase the topojson format size across the board for complex geometries.


nvkelso:
LGTM

zerebubuth:
:+1: for revisiting this and doing the quantisation and clipper fixes for both topojson and MVT. Perhaps we should even do it for all geometry types?

Issue #125, opened 2016-12-01
Memory usage while seeding


hriad:
Hello, 
I'm trying to seed based on a bbox ([-17.628501036130604, 12.780733971799824, 29.499734490797476, 38.04327800816175]) and tiles from zoom level 1 to 16. I'm running this on a 16gb server, but the seeding never completes as tilequeue end up eating all the available memory and finally the server crashes. The queue type is set to file. Is it supposed to consume this much memory?

rmarianski:
The tiles are generated in a lazy fashion, meaning not in memory, but the final step tries to unique everything, which actually does put everything in memory to do so. I bet that if that unique step were omitted, you wouldn't run into memory problems. If you're feeling up to it, you can try removing the [uniquify generator line](https://github.com/tilezen/tilequeue/blob/2d9f041e96159a3ed838678dbef127816682a3e6/tilequeue/command.py#L255) and return back just the `combined_tiles`.

Trying to generate a unique list can be useful as a optimization, but there are cases like yours when you don't want to do it. This should probably be exposed as a configuration option.

By the way, if my back of the envelope math is right, that's about 50 million tiles. That's most likely going to be too much work for a single instance to run through anyway.

rmarianski:
Unfortunately, this is a little more involved than just that one line change, because the caller calls `list` on it to get the count beforehand, which will also realize the entire list into memory.

rmarianski:
I've created #126 and #127 as follow ups.

nvkelso:
@rmarianski should we close this issue in favor of the two new issues you created?

rmarianski:
Sure, #126 specifically should resolve this.

PR #138, opened 2016-12-28
Reduce tilequeue seed memory usage

rmarianski:
Connects to https://github.com/tilezen/tilequeue/issues/125

The seed command was realizing all coordinates in memory to print out the total count before enqueueing. This eliminates that and goes through an intermediary in-memory queue to keep the total memory footprint smaller.

### DIFF
--- a/tilequeue/command.py
+++ b/tilequeue/command.py
@@ -1,3 +1,4 @@
+from collections import Iterable
 from collections import namedtuple
 from contextlib import closing
 from itertools import chain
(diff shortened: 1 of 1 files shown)


iandees:
Can we rewrite the seeder to use a Python generator function instead of passing around a queue? That'd save memory and be a bit more pythonic/cleaner.

rmarianski:
It does use a generator to come up with the list of tiles, but it sticks that onto separate queues for threading purposes. I don't see a good way to do it without queues and keep multi-threading.

iandees:
Gotcha, makes sense 👍.

I think I was thinking of a different bit of code that I was experimenting with when handling the tiles of interest list, but can't find it now.

rmarianski:
> I think I was thinking of a different bit of code that I was experimenting with when handling the tiles of interest list, but can't find it now.

Related to https://github.com/tilezen/tilequeue/pull/122 maybe?

It's very possibly, likely even, that you came across code that could be cleaner :) Feel free to submit a pr to improve it or flag it via an issue to come back to it later.

Issue #139, opened 2017-01-04
Optimize JSON performance via different library


nvkelso:
Profiling tile generation speed with New Relic we've found we spend a good bit of time generating the basic GeoJSON (*.json) tiles – and any TopoJSON (*.topojson) or MVT (*.mvt) tile generated also is required to pass thru the *.json format as a first step.

We think that `import simplejson` in Python (instead of `import json`) will speed things up significantly for all formats. 

Please also include before/after benchmarks.

(Requires additional tileserver work, link to this issue.)

rmarianski:
> and any TopoJSON (.topojson) or MVT (*.mvt) tile generated also is required to pass thru the *.json format as a first step.

To clarify, I don't think it's required, we just do it that way. Circumstances have changed since though, eg we have xonacatl and enqueue the tile coordinate, so I think we should at least re-evaluate whether this is still necessary.

nvkelso:
Let's focus just on `simplejson` configuration change here.

iandees:
I ran some benchmarks this afternoon and came to a surprising finding: Python's built-in `json` is just as fast at what we're doing as `simplejson`. In fact in some cases it's ~1-5% slower. This was surprising to me, but I found [this fairly recent blog post](http://artem.krylysov.com/blog/2015/09/29/benchmark-python-json-libraries/) that benchmark's some Python JSON libraries that gives similar findings.

Their benchmark post includes another, less-featureful library called `ujson`. It gives a fairly sizeable (~10-20%) bump in speed for our topojson and json formats.

My raw numbers are in [this Google Sheet](https://docs.google.com/a/mapzen.com/spreadsheets/d/1RN2ShzRmeV8haCbuvQo9KIrjWf0ePC2TapqWiAhrCiw/edit?usp=sharing) for now, and I'll work on summarizing them in the morning.

nvkelso:
Completed in #143.

PR #143, opened 2017-01-06
Use ujson for serializing GeoJSON and TopoJSON

iandees:
For https://github.com/tilezen/tilequeue/issues/139.

`simplejson` was [merged into Python core in 2.6](https://docs.python.org/3/whatsnew/2.6.html#the-json-module-javascript-object-notation), so there's actually a slow-down using simplejson 3.6.4 (which is currently installed, but unused) or 3.10.0 (the latest version):

| Coord | Format | Python 2.7.6 <br/>Built-in | `simplejson` 3.6.4 | `simplejson` 3.10.0 |
|:--|:--|--:|--:|--:|
| 1/1/1 | json | 358.7ms | 385.3 (+7.42%) | 387.1 (+7.93%) |
| 1/1/1 | topojson | 232.0ms | 252.9ms (+8.98%) | 257.3 (+10.91%) |
| 7/32/47 | json | 897.2 | 910.4ms (+1.46%) |   927.5 (+3.37%) |
| 7/32/47 | topojson | 567.7 | 586.6ms (+3.33%) |   601.3 (+5.92%) |
| 12/1050/1522    | json | 3238.7 |    3366.8ms (+3.95%) |   3368.8 (+4.02%) |
| 12/1050/1522    | topojson | 2128.4 |    2213.2ms (+3.98%) |   2244.2 (+5.44%) |
| 16/16812/24357  | json | 390.2 | 395.9ms (+1.45%) |   412.5 (+5.70%) |
| 16/16812/24357  | topojson | 296.6 | 294.3ms (-0.79%) |  294.7 (-0.63%) |

As a result, this pull request switches to use `ujson`, a faster (but less feature-rich) JSON library. It seems to shave 10-20% of time:

| Coord | Format | Python 2.7.6<br/>Built-in | `ujson` 1.35 |
|:--|:--|--:|--:|
| 1/1/1 | json | 358.7ms | 292.7ms (-18.4%) |
| 1/1/1 | topojson | 232.1ms | 189.1ms (-18.5%) |
| 7/32/47 | json | 897.3ms | 780.9ms (-13%) |
| 7/32/47 | topojson | 567.7ms | 488.9ms (-13.9%) |
| 12/1050/1522 | json | 3238.8ms | 2935ms (-9.4%) |
| 12/1050/1522 | topojson | 2128.5ms | 1858.7ms (-12.7%) |
| 16/16812/24357 | json | 390.3ms | 334.3ms (-14.3%) |
| 16/16812/24357 | topojson | 296.7ms | 240.5ms (-18.9%) |

`ujson` resulted in a < 2% difference for the mvt and mvtb formats.


nvkelso:
Nice work! I was expecting a similar (or slightly smaller) improvement in MVT generation since it first has to calculate the JSON version. Does that mean most the savings here is JSON > String > file IO related changes?

iandees:
My measurements here are for the runtime of the [`_create_formatted_tile()`](https://github.com/tilezen/tilequeue/blob/master/tilequeue/process.py#L250-L266) function for each of the 4 formats, and it doesn't look like the `mvt` format [does anything with JSON at all](https://github.com/tilezen/tilequeue/blob/master/tilequeue/format/mvt.py).

iandees:
@nvkelso said:
> Please also link up related PR for tileserver changes.

A tileserver PR would just be an adjustment of the `tilequeue` dependency, I think. We'd need to merge this PR and tag `tilequeue` first.

nvkelso:
We declare `simplejson` in the dependencies for tileserver now (which seems weird to me, but @rmarianski said it was to play nicer with tilequeue?):

- https://github.com/tilezen/tileserver/blob/master/requirements.txt#L18

And we `import json` in `__init__.py`:

- https://github.com/tilezen/tileserver/blob/master/tileserver/__init__.py#L22

nvkelso:
To observe the expected MVT format times improvement, I suspect you'd additionally need to measure the total time **tileserver** took to generate the example tiles per format (not just the `_create_formatted_tile()` function in tilequeue)?

rmarianski:
> We declare simplejson in the dependencies for tileserver now (which seems weird to me, but @rmarianski said it was to play nicer with tilequeue?):

It should be removed from all requirements files.

> Yep, it looks like tileserver does the JSON decoding stuff here.

Some of that, if not all, was a vestige of when tileserver handled custom layer requests. There's an [issue](https://github.com/tilezen/tileserver/issues/46) for cleaning that up.

iandees:
I checked before and after `ujson` and it looks like the only difference is that the default output for `ujson` is to not include and whitespace between JSON syntax. For example, `ujson` outputs: 

```json
"landuse":{"type":"FeatureCollection","features":[]}}
```

whereas the built-in JSON serializer includes whitespace:

```json
"landuse": {"type": "FeatureCollection", "features": []}}
```

Issue #141, opened 2017-01-04
Add local Postgres connection pooling


nvkelso:
We already do server pooling with PG Bouncer, but we could probably gain further performance by also using local python PG connection pooling. 

This would require some code implementations cleanup (some additions, probably a lot of old funk to remove).

nvkelso:
This was completed in #145.

PR #145, opened 2017-01-19
Use psycopg2 connection pools

iandees:
Connects to https://github.com/tilezen/tilequeue/issues/141

Simplify our code a bit by using psycopg2's built-in connection pools.

### DIFF
--- a/tilequeue/postgresql.py
+++ b/tilequeue/postgresql.py
@@ -1,297 +1,63 @@
 from itertools import chain
 from itertools import cycle
+from psycopg2.pool import ThreadedConnectionPool
 from psycopg2.extras import register_hstore, register_json
 import psycopg2
 import threading
 
 
-class RoundRobinConnectionFactory(object):
+class DatabaseCycleConnectionPool(object):
 
-    def __init__(self, conn_info, hosts):
-        self.conn_info = conn_info
-        self.hosts_gen = iter(cycle(hosts))
+    """
+    Maintains a psycopg2 ThreadedConnectionPool for each of the
+    given dbnames. When a client requests a set of connections,
+    all of those connections will come from the same database.
+    """
 
-    def __call__(self, ignored_dsn):
-        host = self.hosts_gen.next()
-        conn_info = dict(self.conn_info, host=host)
-        conn = psycopg2.connect(**conn_info)
-        return conn
+    def __init__(self, min_conns_per_db, max_conns_per_db,
+            dbnames, conn_info):
+        self._pools = []
+        self._conns_to_pool = {}
 
+        for dbname in dbnames:
+            pool = ThreadedConnectionPool(
+                min_conns_per_db,
+                max_conns_per_db,
+                dbname=dbname,
+                **conn_info,
+            )
+            self._pools.append(pool)
 
-class NoPoolingConnectionPool(object):
+        self._pool_cycle = cycle(self._pools)
+        self._lock = threading.Lock()
 
-    # This adheres to the connection pool interface, but generates a
-    # new connection with every request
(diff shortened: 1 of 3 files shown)


iandees:
The diff is rather large here because I deleted a bunch of code from the `postgresql.py` file when adding the connection pool class.

tilequeue/postgresql.py
```diff
@@ -1,297 +1,66 @@
...
-                        pass
-            self.conns_by_host.clear()
-            self.host_conns_in_use.clear()
-            self.host_conns_not_in_use.clear()
-            self.hosts = []
+                pool_to_use = next(self._pool_cycle)
+                for _ in range(n_conns):
+                    conn = pool_to_use.getconn()
 
+                    register_json(conn, loads=json.loads, globally=True)
```
iandees: A subtle, notable difference here is that I'm passing in `ujson`'s `loads` for `psycopg2` to use when parsing the data from PostgreSQL.

iandees: (This should make things way faster.)

tilequeue/query.py
```diff
@@ -157,16 +157,17 @@ def __init__(self, conn_info, layer_data, io_pool, n_conn):
 
         self.dbnames = self.conn_info.pop('dbnames')
         self.dbnames_query_index = 0
-        self.sql_conn_pool = DBAffinityConnectionsNoLimit(
-            self.dbnames, n_conn, self.conn_info)
+        self.sql_conn_pool = DatabaseCycleConnectionPool(
+            n_conn, 100, self.dbnames, self.conn_info)
```
iandees: I have a hard-coded 100 connection maximum here but it should probably be a configurable thing. The only time this gets checked is when we're requesting [a new connection out of the pool](https://github.com/psycopg/psycopg2/blob/master/lib/pool.py#L90-L91). The connection pools will only maintain `n_conn` connections per `dbname` otherwise they'll get closed whenever the connection is returned to the pool.

nvkelso: How many active connections do we maintain per DB now?

How many are typically from a single tileserver or tilequeue instance / worker?

rmarianski: I wouldn't expect us to ever need to go over the minimum. What we can use for the maximum is the conservative number where we calculate how many io threads we'll need: https://github.com/tilezen/tilequeue/blob/738038db62b9cd33b35b4d9377c5c41d00625e99/tilequeue/command.py#L528

We can just pass that on through to the datafetcher constructor.

tilequeue/postgresql.py
```diff
@@ -1,297 +1,66 @@
...
-        for conn in conns:
-            try:
-                conn.close()
-            except:
-                pass
+        with self._lock:
+            for conn in conns:
+                pool = self._conns_to_pool.pop(id(conn), None)
+                if pool:
+                    pool.putconn(conn)
```
rmarianski: I recognize that this case should never happen, but should we have an else clause where we try to close the connection? Actually an assert might make more sense because this means a caller tried to close a connection that it didn't get from a pool.

iandees: Yep, good idea.

iandees: Looking closer: I didn't use an assert because I didn't want to dump out of the loop and leave the rest of the connections in the `conns` list open.

iandees: … but I'll switch to assert so we just dump out of here in that case. It's not particularly recoverable with existing code anyway.

rmarianski: I see what you're saying. If we cared, we could try to close all the connections that were valid before raising, but I agree that this should trigger a failure case anyway.

PR #150, opened 2017-02-02
Track connections before trying to use them

iandees:
This way we can give them back cleanly if any errors happen while trying to use them.


iandees:
It seems to have stopped them in my test last night. I only changed one tilequeue and one tileserver in situ though, so hopefully deploying this change across all hosts on dev will show that it works.

Issue #152, opened 2017-02-13
Write-tile-if-changed should account for zipfile metadata


zerebubuth:
The `store.write_tile_if_changed` function tries to reduce the number of writes to the store by fetching any existing tile at that coordinate from the store and only writing the new tile if they differ. Formats such as JSON or MVT contain only the data about the content of the tile, and metadata such as the file timestamp is transmitted in the HTTP headers, which aren't used in the test of whether the tiles differ.

The `zipfile` format used by the metatiles stores such metadata internally, meaning that if two tiles are generated at different times then they will differ, even if the tile contents are identical. This prevents the `write_tile_if_changed` optimisation from working.

Instead, we should add a special case for comparison of two zip files which compares only the set of tiles in the file and their contents.

zerebubuth:
Fixed in #153.

PR #153, opened 2017-02-13
Compare ZIP file ignoring metadata

zerebubuth:
When comparing zip files to see if the tile already in storage is the same as the one just generated, ignore the timestamp and archive file order, as these are likely to change between two renders. Instead, all we care about is that they contain the same set of files and that those files have identical content. 

Connects to #152.

@rmarianski, @iandees could you review, please?

### DIFF
--- a/tests/test_metatile.py
+++ b/tests/test_metatile.py
@@ -79,7 +79,7 @@ class TestMetatile(unittest.TestCase):
         self.assertEqual(json, extracted)
 
     def test_metatile_file_timing(self):
-        from time import sleep, gmtime, time
+        from time import gmtime, time
         from tilequeue.metatile import metatiles_are_equal
 
         # tilequeue's "GET before PUT" optimisation relies on being able to
--- a/tests/test_metatile.py
+++ b/tests/test_metatile.py
@@ -77,3 +77,27 @@ class TestMetatile(unittest.TestCase):
         buf = StringIO.StringIO(metatiles[0]['tile'])
         extracted = extract_metatile(1, buf, tile)
         self.assertEqual(json, extracted)
+
+    def test_metatile_file_timing(self):
+        from time import sleep, gmtime, time
+        from tilequeue.metatile import metatiles_are_equal
+
+        # tilequeue's "GET before PUT" optimisation relies on being able to
+        # fetch a tile from S3 and compare it to the one that was just
+        # generated. to do this, we should try to make the tiles as similar
+        # as possible across multiple runs.
+
+        json = "{\"json\":true}"
+        tiles = [dict(tile=json, coord=Coordinate(0, 0, 0),
+                      format=json_format, layer='all')]
+
+        when_will_then_be_now = 10
+        t = time()
+        now = gmtime(t)[0:6]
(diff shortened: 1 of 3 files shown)


tests/test_metatile.py
```diff
@@ -77,3 +77,27 @@ def test_extract_metatiles_single(self):
...
+        # tilequeue's "GET before PUT" optimisation relies on being able to
+        # fetch a tile from S3 and compare it to the one that was just
+        # generated. to do this, we should try to make the tiles as similar
+        # as possible across multiple runs.
+
+        json = "{\"json\":true}"
+        tiles = [dict(tile=json, coord=Coordinate(0, 0, 0),
+                      format=json_format, layer='all')]
+
+        when_will_then_be_now = 10
```
rmarianski: great variable name! :zap:

tilequeue/metatile.py
```diff
@@ -16,6 +18,10 @@ def make_single_metatile(size, tiles):
     if len(tiles) == 0:
         return []
 
+    if date_time is None:
+        from time import gmtime
```
rmarianski: move import top?

tilequeue/store.py
```diff
@@ -144,6 +144,24 @@ def make_s3_store(bucket_name,
...
 
+def tiles_are_equal(tile_data_1, tile_data_2, fmt):
+    """
+    Returns True if the tile data is equal in tile_data_1 and tile_data_2. For
+    most formats, this is a simple byte-wise equality check. For zipped
+    metatiles, we need to check the contents, as the zip format includes
+    metadata such as timestamps and doesn't control file ordering.
+    """
+
+    from tilequeue.format import zip_format
```
rmarianski: move import top?

tilequeue/store.py
```diff
@@ -144,6 +144,24 @@ def make_s3_store(bucket_name,
...
+
+    if fmt and fmt == zip_format:
+        from tilequeue.metatile import metatiles_are_equal
```
rmarianski: move import top?

tilequeue/metatile.py
zerebubuth: Good point! I had picked up the wrong habit from the tests. And today I learned that there's a Python style thing that Flake8 _doesn't_ complain about :wink:

zerebubuth:
Moved imports to top and used `cStringIO` in 90850f0,

tilequeue/metatile.py
```diff
@@ -1,13 +1,16 @@
 import zipfile
-import StringIO
+import cStringIO as StringIO
```
iandees: ❤️

tilequeue/metatile.py
rmarianski: :laughing: If we want to get more opinionated about imports, I personally like to have just 1 import  per line and sort them alphabetically. 1 import per line makes it easier to search for, and sorting alphabetically just takes all the thinking out of it. But, I'm with you that I'm happy the hateful flake8 doesn't complain about it ;)

tilequeue/metatile.py
rmarianski: :+1: This is even better than what I was suggesting.

PR #154, opened 2017-02-13
Return a better error for invalid dates

rmarianski:
When we get invalid dates, we can have the edtf date api return back `None` for earliest/latest dates. This adds a check to guard against that, and return back a better error message in this case. Otherwise, the `min` or something else downstream will raise a general error when it's first used.

### DIFF
--- a/tilequeue/wof.py
+++ b/tilequeue/wof.py
@@ -327,13 +327,26 @@ def create_neighbourhood_from_json(json_data, neighbourhood_meta):
     edtf_cessation = _normalize_edtf(props.get('edtf:cessation'))
     edtf_deprecated = _normalize_edtf(props.get('edtf:deprecated'))
 
+    # check that the dates are valid first to return back a better error
+    inception_earliest = edtf_inception.date_earliest()
+    cessation_latest = edtf_cessation.date_latest()
+    deprecated_latest = edtf_deprecated.date_latest()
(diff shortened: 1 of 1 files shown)


rmarianski:
@zerebubuth, @iandees can you review please?

tilequeue/wof.py
```diff
@@ -327,13 +327,24 @@ def failure(reason):
     edtf_cessation = _normalize_edtf(props.get('edtf:cessation'))
     edtf_deprecated = _normalize_edtf(props.get('edtf:deprecated'))
 
+    # check that the dates are valid first to return back a better error
+    inception_earliest = edtf_inception.date_earliest()
+    cessation_latest = edtf_cessation.date_latest()
+    deprecated_latest = edtf_deprecated.date_latest()
+    if inception_earliest is None:
+        return failure('invalid edtf:inception: %s' % props['edtf:inception'])
```
zerebubuth: It's possible for `props` not to have an `edtf:...` key, which is hidden by `_normalize_edtf` returning a valid `EDTF` object when given `None`. This might cause a `KeyError` here, though?

Perhaps `failure('invalid edtf:inception: %r' % props.get('edtf:inception'))`?

rmarianski: Well spotted! Fixed in f133a5d.

PR #160, opened 2017-02-17
``tilequeue.store.TileDirectory`` must write files *atomically*

antonov-impulsm:
``tilequeue.store.TileDirectory`` must write files *atomicly* ..

problem description
-------------------------

* if server process will stopped suddenly -- cached must be protected from corruption. ``Gunicorn`` may kill process (any time) if timer of timeout of request elapsed

* cached must not able to read when it is not yet complete written.  ``Gunicorn`` runs several concurrent processes with one store directory.

patch details
----------------

introduced function ``os_replace(src, dst)`` for *unix-style* atomically file writing

### DIFF
--- a/tilequeue/store.py
+++ b/tilequeue/store.py
@@ -2,10 +2,14 @@
 
 from boto import connect_s3
 from boto.s3.bucket import Bucket
+from builtins import range
+from future.utils import raise_from
 import md5
 import os
 from tilequeue.metatile import metatiles_are_equal
 from tilequeue.format import zip_format
+import random
+import threading
 
 
 def calc_hash(s):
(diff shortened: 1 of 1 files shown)


tilequeue/store.py
```diff
@@ -100,11 +164,22 @@ def write_tile(self, tile_data, coord, format, layer):
...
+            os.getpid(),
+            threading.currentThread().ident,
+            random.randint(1,1000000)
+        )
+
+        with open(swap_file_path, 'w') as tile_fp:
             tile_fp.write(tile_data)
 
+        # write file as atomic operation
+        os_replace(swap_file_path, file_path)
```
rmarianski: Do you care about trying to remove the temp file if there was an error?

antonov-impulsm:
> Do you care about trying to remove the temp file if there was an error?

thank you for remark!

I added commit (72869f9) -- to using ``try-finally`` (not ``try-catch``) for remove swap-file in error cases.

I think it is *may be* useful (but I not *fully* confident about *real* use-cases it). please, give your opinion. thank you

rmarianski:
My only reservation would be that you're paying for the cost of the exists check, even in the more common case that everything is successful. Do you think it's safe enough to only perform this check if the rename raised an exception? On success, I'd expect us to be confident that the temp file was removed as part of the process.

In other words, I was thinking something along the lines of the below (untested):

```python
with open(swap_file_path, 'w') as tile_fp:
    tile_fp.write(tile_data)

try:
    # write file as atomic operation
    os_replace(swap_file_path, file_path)
except Exception as e:
    try:
        os.remove(swap_file_path)
    except:
        pass
    raise e
```

But if you don't think the exists check is ever going to be a source of slow down, then it's certainly safer the way you have it written.

antonov-impulsm:
> My only reservation would be that you're paying for the cost of the exists check, even in the more common case that everything is successful.

I agree, thank you

rmarianski:
Thanks for the pr!

PR #164, opened 2017-03-07
Extend metatile extract code to work with 2x2 metatiles

zerebubuth:


zerebubuth:
Looks like more appdirs-related problems in the CI system :disappointed:

PR #165, opened 2017-03-07
Emit additional metrics during intersection

rmarianski:
Connects to https://github.com/mapzen/tile-tasks/issues/201
Refs https://github.com/mapzen/chef-mapzen_tilequeue/pull/29

This emits the stats to the log, and logstash will pick them up and send them over to statsd.

Longer term I think we should transition to the same approach we took in tapalcatl, namely to have the process itself send the stats directly.

### DIFF
--- a/tilequeue/command.py
+++ b/tilequeue/command.py
@@ -272,31 +272,64 @@ def tilequeue_drain(cfg, peripherals):
 
 
 def explode_and_intersect(coord_ints, tiles_of_interest, until=0):
+
     next_coord_ints = coord_ints
     coord_ints_at_parent_zoom = set()
+
+    total_coord_ints = []
+
+    # to capture metrics
+    total = 0
+    hits = 0
+    misses = 0
+
     while True:
+
         for coord_int in next_coord_ints:
+
+            total += 1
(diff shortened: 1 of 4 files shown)


tilequeue/command.py
```diff
@@ -271,32 +271,60 @@ def tilequeue_drain(cfg, peripherals):
...
+            else:
+                misses += 1
+
             zoom = zoom_mask & coord_int
             if zoom > until:
                 parent_coord_int = coord_int_zoom_up(coord_int)
                 coord_ints_at_parent_zoom.add(parent_coord_int)
+
         if not coord_ints_at_parent_zoom:
+            intersect_metrics['total'] = total
```
iandees: Modifying a `dict` that's passed in to the function is not very pythonic and could lead to unintended side-effects. It looks like the `coord_ints` thing we're passing in here is just a `set` of `set`s (i.e. not a generator or iterator) so we don't save anything by using a generator here. Can we just return the list of `coord_int` instead of `yield`ing them individually and also return the stats object?

rmarianski: Fair point, updated in 7ddf5ef.

PR #166, opened 2017-03-07
Compress files in metatile.

zerebubuth:
On a random tile that I was testing with, this reduced file size by 79%.

The compressed metatile will take up less space in storage and transfer more quickly over the network. The disadvantage would be the CPU time required to decompress it. Looking at CPU usage on the tapalcatl machines, it's fairly light at the moment, so we've room to do more there.

What do you think? Should I make it configurable in case we need to turn it off in a hurry?


zerebubuth:
> I thought it was already compressed :)

I know! I thought that too, but the docs clearly say that `ZIP_STORED` is the default, so I should pay more attention.

What they don't mention is that it can be set either in the `ZipFile` constructor, or in the `writestr` method, and that the default in `writestr` _overrides_ an explicit setting in `ZipFile`'s constructor. That one had me confused for a long time.

PR #175, opened 2017-03-20
Fix LinearRing error

zerebubuth:
Was seeing this error in the logs. It appears to be because a MultiPolygon is being interpreted as a Polygon, and therefore has the wrong number of outer rings. To fix this, we just decompose the MultiPolygon into its constituent Polygons.


zerebubuth:
Thanks! Good point about the comment - added one in cc9cc68.

PR #176, opened 2017-03-20
Command to automatically prune tiles of interest

iandees:
Adding a command that prunes the tiles of interest by checking for tiles that were frequently requested and removing tiles from the tiles of interest set (and from S3) that **were not** frequently requested.

### DIFF
--- a/tilequeue/command.py
+++ b/tilequeue/command.py
@@ -965,19 +965,23 @@ def tilequeue_prune_tiles_of_interest(cfg, peripherals):
     assert redshift_uri, ("A redshift connection URI must "
                           "be present in the config yaml")
 
+    redshift_days_to_query = cfg.yml.get('redshift_days')
+    assert redshift_days_to_query, ("Number of days to query "
+                                    "redshift is not specified")
+
     tiles_recently_requested = set()
     with psycopg2.connect(redshift_uri) as conn:
         with conn.cursor() as cur:
             cur.execute("""
                 select x, y, z
                 from tile_traffic_v4
-                where (date >= dateadd(day, -30, current_date))
+                where (date >= dateadd(day, -{days}, current_date))
                   and (z between 0 and 16)
                   and (x between 0 and pow(2,z)-1)
                   and (y between 0 and pow(2,z)-1)
                 group by z, x, y
                 order by z, x, y
-            );""")
+            );""".format(days=redshift_days_to_query))
             n_trr = cur.rowcount
             for (x, y, z) in cur:
                 coord = create_coord(x, y, z)
--- a/tilequeue/command.py
+++ b/tilequeue/command.py
@@ -956,17 +956,17 @@ def tilequeue_prune_tiles_of_interest(cfg, peripherals):
     logger.info('Fetching tiles recently requested ...')
     import psycopg2
 
-    redshift = cfg.yml.get('redshift', {})
+    prune_cfg = cfg.yml.get('toi-prune', {})
 
-    redshift_uri = redshift.get('uri')
+    redshift_uri = prune_cfg.get('redshift-uri')
     assert redshift_uri, ("A redshift connection URI must "
                           "be present in the config yaml")
 
-    redshift_days_to_query = redshift.get('days')
+    redshift_days_to_query = prune_cfg.get('days')
     assert redshift_days_to_query, ("Number of days to query "
                                     "redshift is not specified")
 
-    tiles_recently_requested = set()
+    new_toi = set()
     with psycopg2.connect(redshift_uri) as conn:
         with conn.cursor() as cur:
             cur.execute("""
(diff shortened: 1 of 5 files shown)


tilequeue/command.py
```diff
@@ -948,6 +949,65 @@ def tilequeue_enqueue_tiles_of_interest(cfg, peripherals):
...
+    assert redshift_uri, ("A redshift connection URI must "
+                          "be present in the config yaml")
+
+    tiles_recently_requested = set()
+    with psycopg2.connect(redshift_uri) as conn:
+        with conn.cursor() as cur:
+            cur.execute("""
+                select x, y, z
+                from tile_traffic_v4
+                where (date >= dateadd(day, -30, current_date))
```
rmarianski: Do we want to make this time window configurable?

tilequeue/command.py
```diff
@@ -948,6 +949,65 @@ def tilequeue_enqueue_tiles_of_interest(cfg, peripherals):
...
+        # Remove from the redis toi set
+        peripherals.redis_cache_index.remove_tile_of_interest(coord_int)
+
+        # Remove the tile from S3
+
+        pass
+
+    for coord_int in toi_to_remove:
+        # FIXME: Think about doing this in a thread/process pool
+        delete_tile_of_interest(coord_int)
```
rmarianski: We can also think about formalizing this a bit more and doing this out of process. What's the order of the amount that we've been managing in the past, several million?

tilequeue/tile.py
```diff
@@ -33,6 +33,10 @@ def deserialize_coord(coord_string):
     return coord
 
 
+def create_coord(x, y, z):
+    return Coordinate(row=x, column=y, zoom=z)
```
rmarianski: column and row transposed

```python
Coordinate(column=x, row=y, zoom=z)
```

tilequeue/command.py
iandees: I haven't calculated the number of tiles that would be removed. I can do that now.

I was thinking about putting together an SQS queue and a worker process to do the deletes, but that seemed heavy-handed. Maybe a lambda task to do the delete?

rmarianski: > I was thinking about putting together an SQS queue and a worker process to do the deletes, but that seemed heavy-handed. Maybe a lambda task to do the delete?

I was thinking the same. It's operationally heavier, but I think we'll need something like that if we want to scale past multiple processes/threads on a single instance.

Maybe a good use case for [batch](http://docs.aws.amazon.com/batch/latest/userguide/what-is-batch.html)?

tilequeue/tile.py
iandees: Fixed in 254ac0f.

tilequeue/command.py
iandees: Fixed in 071cc52.

tilequeue/command.py
rmarianski: Not sure if this is a [comprehensive list](http://docs.aws.amazon.com/lambda/latest/dg/use-cases.html) of event sources for lambda, but thinking about it more I think that's a reasonable option. One idea is that we can split up the list into groups of 10k or so, push those groups to a location on s3, and have lambda listen to that. Lambda would perform the delete and remove that object from s3.

tilequeue/command.py
```diff
@@ -953,49 +953,71 @@ def tilequeue_prune_tiles_of_interest(cfg, peripherals):
...
-    tiles_recently_requested = set()
+    new_toi = set()
     with psycopg2.connect(redshift_uri) as conn:
         with conn.cursor() as cur:
             cur.execute("""
                 select x, y, z
                 from tile_traffic_v4
                 where (date >= dateadd(day, -{days}, current_date))
-                  and (z between 0 and 16)
+                  and (z between 10 and 16)
```
nvkelso: should `16` here be a configurable **max zoom**? Right now that's 16, but with 2x2 metatiles it'd be 15.

tilequeue/command.py
iandees: I just tested this on dev and each run of the `delete_tile_of_interest()` function (which deletes 1000 tiles at a time) takes ~2 seconds. With ~9 million tiles to remove, that'll take ~5 hours or so. Subsequent runs should be faster.

iandees:
Here's a (redacted) bit of yaml that I added to my tilequeue config.yaml file to run this on dev.

```yaml
toi-prune:
  redshift-uri: postgresql://user:pass@localhost:5439/analytics
  days: 30
  s3:
    bucket: mapzen-tiles-dev
    date-prefix: 20170123
    path: osm
    layer: all
    format: zip
  always-include-bboxes:
    conus:
      bbox: -124.8,24.8,-66.1,49.3
      min_zoom: 11
      max_zoom: 14
    world:
      bbox: -180.0,-85.06,180.0,85.06
      min_zoom: 0
      max_zoom: 10
```

iandees:
Going to merge this for now and will work on supporting the 512/256 hybrid stuff later.

PR #178, opened 2017-03-28
Add the ability to list individual tiles that should stay in the TOI

iandees:
Follows #176, adds the ability to list individual tiles that should stay in the TOI

### DIFF
--- a/tilequeue/command.py
+++ b/tilequeue/command.py
@@ -994,17 +994,23 @@ def tilequeue_prune_tiles_of_interest(cfg, peripherals):
 
     logger.info('Fetching tiles recently requested ... done. %s found', n_trr)
 
-    for name, info in prune_cfg.get('always-include-bboxes', {}).items():
+    for name, info in prune_cfg.get('always-include', {}).items():
         logger.info('Adding in tiles from %s ...', name)
 
-        bounds = map(float, info['bbox'].split(','))
-        bounds_tileset = set()
-        for coord in tile_generator_for_single_bounds(
-                        bounds, info['min_zoom'], info['max_zoom']):
-            coord_int = coord_marshall_int(coord)
-            bounds_tileset.add(coord_int)
-        n_inc = len(bounds_tileset)
-        new_toi = new_toi.union(bounds_tileset)
+        immortal_tiles = set()
+        if 'bbox' in info:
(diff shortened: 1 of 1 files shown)


iandees:
Sample config:

```yaml
toi-prune:
  redshift-uri: postgresql://redacted:redacted@localhost:5439/analytics
  days: 30
  s3:
    bucket: mapzen-tiles-dev-us-east
    date-prefix: 20170123
    path: osm
    layer: all
    format: zip
  always-include:
    conus:
      bbox: -124.8,24.8,-66.1,49.3
      min_zoom: 10
      max_zoom: 13
    world:
      bbox: -180.0,-85.06,180.0,85.06
      min_zoom: 0
      max_zoom: 9
    tests:
      tiles: [5/2/4, 5/7/8]
```

zerebubuth:
Nice!

tilequeue/command.py
```diff
@@ -994,17 +994,23 @@ def tilequeue_prune_tiles_of_interest(cfg, peripherals):
...
+        if 'bbox' in info:
+            bounds = map(float, info['bbox'].split(','))
+            for coord in tile_generator_for_single_bounds(
+                            bounds, info['min_zoom'], info['max_zoom']):
+                coord_int = coord_marshall_int(coord)
+                immortal_tiles.add(coord_int)
+        elif 'tiles' in info:
+            tiles = map(deserialize_coord, info['tiles'])
+            tiles = map(coord_marshall_int, tiles)
+            immortal_tiles.update(tiles)
```
rmarianski: Can I make a feature request? :) How about adding something like `path` or `file(s)`. We would expect that the file contains coordinates, one per line.

I'm expecting that when we add the tests, what we'll do is first run the script that prints them out. And then we can save that to a file and feed it into this. Maybe another option is to read it from stdin to avoid having the temporary file?

iandees: Done in e17b902.

nvkelso:
Is there an assumption that the list of tiles is always 512px (2x2)
metatile size?

iandees:
Yea, good point. The tiles at this point are dumped straight into the TOI so they are assumed to be 512px.

tilequeue/command.py
rmarianski: :+1:

PR #179, opened 2017-03-29
Account for tilesize on RedShift records

iandees:
Follow of #178, #176.

Standardize on 512px-sized tiles in the tiles of interest by "zooming out" the records from RedShift if they don't specify 512px as the requested tile size.

### DIFF
--- a/tilequeue/command.py
+++ b/tilequeue/command.py
@@ -981,7 +981,7 @@ def tilequeue_prune_tiles_of_interest(cfg, peripherals):
                 select x, y, z, tilesize
                 from tile_traffic_v4
                 where (date >= dateadd(day, -{days}, current_date))
-                  and (z between 10 and 16)
+                  and (z between 0 and 15)
                   and (x between 0 and pow(2,z)-1)
                   and (y between 0 and pow(2,z)-1)
                 group by z, x, y
--- a/tilequeue/command.py
+++ b/tilequeue/command.py
@@ -978,7 +978,7 @@ def tilequeue_prune_tiles_of_interest(cfg, peripherals):
     with psycopg2.connect(redshift_uri) as conn:
         with conn.cursor() as cur:
             cur.execute("""
-                select x, y, z
+                select x, y, z, tilesize
                 from tile_traffic_v4
                 where (date >= dateadd(day, -{days}, current_date))
                   and (z between 10 and 16)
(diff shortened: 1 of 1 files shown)


tilequeue/command.py
```diff
@@ -978,18 +978,23 @@ def tilequeue_prune_tiles_of_interest(cfg, peripherals):
...
                 where (date >= dateadd(day, -{days}, current_date))
-                  and (z between 10 and 16)
+                  and (z between 0 and 15)
                   and (x between 0 and pow(2,z)-1)
                   and (y between 0 and pow(2,z)-1)
                 group by z, x, y
                 order by z, x, y;""".format(days=redshift_days_to_query))
             n_trr = cur.rowcount
-            for (x, y, z) in cur:
+            for (x, y, z, tile_size) in cur:
```
rmarianski: Don't need the parens here, but I know that @zerebubuth likes seeing them so up to you :)

tilequeue/command.py
```diff
@@ -978,18 +978,23 @@ def tilequeue_prune_tiles_of_interest(cfg, peripherals):
     with psycopg2.connect(redshift_uri) as conn:
         with conn.cursor() as cur:
             cur.execute("""
-                select x, y, z
+                select x, y, z, tilesize
                 from tile_traffic_v4
                 where (date >= dateadd(day, -{days}, current_date))
-                  and (z between 10 and 16)
+                  and (z between 0 and 15)
```
rmarianski: Are we going to be pruning low zoom level tiles too? ie 0-10, err 0-9 now?

tilequeue/command.py
zerebubuth: Indeed! I do like parens - mainly because I can't remember precedence rules or, as in this case, like being reminded that we're destructuring a tuple.

tilequeue/command.py
iandees: I made this change because I didn't want to assume we always wanted to have zooms 0 through 9 in the TOI. In my sample configuration I have the whole Earth at z0-z9 as an "immortal" area, so it gets added to the TOI that way.

tilequeue/command.py
iandees: I was being extra explicit with parens in the SQL above so I kept going in the Python here. 😄 I'm going to keep it.

PR #180, opened 2017-03-31
Add to tiles of interest as part of tilequeue "pruner" command

iandees:
This adds the ability to include new tiles in the tiles of interest as part of the "prune" tilequeue command that was added in #176. This is a counterpart to https://github.com/tilezen/tileserver/pull/85, which pulls out the enqueueing and TOI adding piece from tileserver.

### DIFF
--- a/tilequeue/command.py
+++ b/tilequeue/command.py
@@ -1077,20 +1077,19 @@ def tilequeue_prune_tiles_of_interest(cfg, peripherals):
     logger.info('Computing tiles to add ... done. %s found',
                 len(toi_to_add))
 
-    logger.info('Adding %s tiles to TOI and SQS ...',
+    logger.info('Enqueueing %s tiles to SQS ...',
                 len(toi_to_add))
 
-    def add_tile_of_interest(coord_int):
-        # Add to Redis
+    sqs_queue = peripherals.queue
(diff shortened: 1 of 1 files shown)


tilequeue/command.py
```diff
@@ -1075,6 +1072,26 @@ def delete_tile_of_interest(s3_parts, coord_ints):
...
+                len(toi_to_add))
+
+    def add_tile_of_interest(coord_int):
+        # Add to Redis
+
+        # Enqueue a render request for the tile
+        pass
+
+    for coord_int in toi_to_add:
+        add_tile_of_interest(coord_int)
```
rmarianski: Flow looks good to me. Are we going to be enqueueing the coordinates with custom priorities? Or like normal based on zoom? If normal, there are some existing utilities that should make it easier.

iandees: I was planning on using the normal priority queues.

iandees: I'm planning on using the existing priority queues.

rmarianski: Ok, you can probably look at the [enqueue tiles of interest command](https://github.com/tilezen/tilequeue/blob/9fa942f41adad206fabe473a693207a9548eb17d/tilequeue/command.py#L932-L952) for inspiration.

PR #181, opened 2017-04-03
Move the zoom to enqueue to 15

rmarianski:
This gets run when enqueueing the tiles of interest, when we deploy to prod and enqueue all the tiles. This should be set to z15 now.

This wasn't changing really changing before, but maybe this should be set from configuration now? Or perhaps we do it next time if/when this changes?

### DIFF
--- a/config.yaml.sample
+++ b/config.yaml.sample
@@ -84,8 +84,8 @@ tiles:
     # process on update
     parent-zoom-until: 11
 
-  # this is the max zoom used when enqueueing the tiles of interest
-  enqueue_max_zoom: 15
+  # zoom where tile content stops changing
+  max-zoom-with-changes: 16
 
 process:
   # number of simultaneous "querysets" to issue to the database.  The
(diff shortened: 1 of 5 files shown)


zerebubuth:
:+1: for making this configurable, too.

rmarianski:
Made the zoom to enqueue configurable in c6bf885. But in the future, we might want to see if we can merge it in together with https://github.com/tilezen/tilequeue/blob/9fa942f41adad206fabe473a693207a9548eb17d/tilequeue/worker.py#L108. At that point, maybe it should really be set to 16, and we subtract away the metatile zoom just like that code in the worker does?

zerebubuth:
> At that point, maybe it should really be set to 16, and we subtract away the metatile zoom just like that code in the worker does?

Yup, exactly. That way we could change the metatile zoom without having to remember that it's coupled to the max zoom.

rmarianski:
> At that point, maybe it should really be set to 16, and we subtract away the metatile zoom just like that code in the worker does?
> 
> Yup, exactly. That way we could change the metatile zoom without having to remember that it's coupled to the max zoom.

Updated in cb38515.

PR #182, opened 2017-04-06
A couple cleanups from running the gardener locally

iandees:
A couple cleanups from #180.

- Have to include tilesize in the redshift group by query
- Filter out null tile coords that might sneak in to the immortal tiles section
- The redis client expects srem args to be a star-arg, not a list

### DIFF
--- a/tilequeue/cache/redis_cache_index.py
+++ b/tilequeue/cache/redis_cache_index.py
@@ -18,6 +18,8 @@ class RedisCacheIndex(object):
                 yield coord
 
     def remove_tiles_of_interest(self, coord_ints):
+        # Note that this will end up as a single SREM call to Redis,
+        # so you should probably group the coords into ~1000s.
         return self.redis_client.srem(self.cache_set_key, *coord_ints)
 
     def fetch_tiles_of_interest(self):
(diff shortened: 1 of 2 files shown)


tilequeue/cache/redis_cache_index.py
```diff
@@ -18,7 +18,7 @@ def intersect(self, coords, tiles_of_interest=None):
                 yield coord
 
     def remove_tiles_of_interest(self, coord_ints):
-        return self.redis_client.srem(self.cache_set_key, coord_ints)
+        return self.redis_client.srem(self.cache_set_key, *coord_ints)
```
rmarianski: Nice catch! :zap: 

This will turn into a single call to redis, so we probably shouldn't call this with millions of integers. (Not sure if we are or aren't, just caught my eye).

iandees: It's currently [getting called in groups of 1000](https://github.com/tilezen/tilequeue/blob/c8160d77b4ab859d10e1e8854319238bfe2f784b/tilequeue/command.py#L1073-L1074). It's probably a good thing to document, though.

rmarianski: :+1: I was just lazy and didn't check :)

PR #184, opened 2017-04-10
Add configurable request and tile count cutoffs

iandees:
This allows us to specify a maximum number of tiles to include from RedShift in the tiles of interest (the "immortal tiles" will always be included). It also adds a tile request count cutoff. These two cutoff values should let us more carefully specify how much work we want tilequeue to do.

### DIFF
--- a/tilequeue/command.py
+++ b/tilequeue/command.py
@@ -1,5 +1,6 @@
 from collections import Iterable
 from collections import namedtuple
+from collections import defaultdict
 from contextlib import closing
 from itertools import chain
 from jinja2 import Environment
@@ -977,7 +978,7 @@ def tilequeue_prune_tiles_of_interest(cfg, peripherals):
     assert s3_parts, ("The name of an S3 bucket containing tiles "
                       "to delete must be specified")
 
-    redshift_results = dict()
+    redshift_results = defaultdict(int)
     with psycopg2.connect(redshift_uri) as conn:
         with conn.cursor() as cur:
             cur.execute("""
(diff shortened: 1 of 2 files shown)


tilequeue/command.py
```diff
@@ -962,43 +962,84 @@ def tilequeue_prune_tiles_of_interest(cfg, peripherals):
...
-                new_toi.add(coord_int)
+                # Sum the counts from the 256 and 512 tile requests into the
+                # slot for the 512 tile.
+                existing_count = redshift_results.get(coord_int)
+                if existing_count:
+                    existing_count += count
+                else:
+                    existing_count = count
+
+                redshift_results[coord_int] = existing_count
```
rmarianski: If you use a `defaultdict(int)`, this can become: `redshift_results[coord_int] += count`

tilequeue/command.py
```diff
@@ -962,43 +962,84 @@ def tilequeue_prune_tiles_of_interest(cfg, peripherals):
...
+
+    # Sort redshift_results (a dict of coord_int->score) by the score value,
+    # filter out the coords that didn't meet the requests cutoff,
+    # then take the first `cutoff_tiles`.
+
+    new_toi = set(
+        t[0] for t in filter(
+            lambda t: t[1] >= cutoff_requests,
+            sorted(redshift_results.iteritems(), key=lambda t: t[1])
+        )[:cutoff_tiles]
```
rmarianski: Do you think this would be easier to read if it were spelled out more? Something like: (untested)

```python
new_toi = set()
for coord_int, score in sorted(redshift_results.iteritems(), key=operator.itemgetter(1), reverse=True):
    if score >= cutoff_requests:
        new_toi.add(coord_int)
        if len(new_toi) >= cutoff_tiles:
            break
```

iandees: Yea, apparently I was feeling a little over-optimiz-ey last night. I'll make that clearer.

PR #189, opened 2017-04-17
Use getdate() instead of current_date in RedShift query

iandees:
current_date returns only the date, not the whole datetime. getdate() returns the current datetime.


iandees:
(See [docs for `getdate()`](http://docs.aws.amazon.com/redshift/latest/dg/r_GETDATE.html), which calls out that the parens are required.

PR #193, opened 2017-04-25
Move tiles of interest set out of Redis and into S3

iandees:
Moves the tiles of interest set out of Redis and into S3.


tilequeue/command.py
```diff
@@ -874,59 +874,48 @@ def tilequeue_seed(cfg, peripherals):
...
-        redis_queue.put(None)
-        thread_redis.join()
-        logger.info('Tiles of interest ... done')
 
     thread_enqueue.join()
-    logger.info('Sqs ... done')
+    logger.info('Enqueueing to SQS ... done')
+
+    if cfg.seed_should_add_to_tiles_of_interest:
+        logger.info('Adding to Redis ... ')
```
nvkelso: Per the PR title, is Redis still needed here?

nvkelso:
Do we want to allow read/write to local disk for the TOI (not just S3)?

iandees:
This isn't ready for review yet. I still have lots to go :) These (and other things that Rob brought up with me) are all good points.

nvkelso:
Woot! I've added `WIP: ` to the PR to make that clear.

tilequeue/command.py
iandees: This was a terminology problem (should have said "Tiles of Interest" not "Redis").

Fixed in 7a52e76

iandees:
This is just about done, with the exception of a couple things:

- [ ] Tests are failing: It's not clear to me how the configuration tests work with an empty config. if [`self.yml['toi-store']`](https://github.com/tilezen/tilequeue/blob/cf1e28db8db5fcdb9ac5e58e053b430d181c5401/tilequeue/config.py#L44) fails because of a `KeyError`, why doesn't [`self.yml['tiles']['seed']`](https://github.com/tilezen/tilequeue/blob/cf1e28db8db5fcdb9ac5e58e053b430d181c5401/tilequeue/config.py#L28) also fail right above it? Is there a default config that's getting used somewhere?
- [ ] The serializing/deserializing code assumes a tiles of interest set that contains `Coordinate` objects, whereas it contained `coord_int`s before. Since we're storing as strings anyway, I thought it'd be nicer to end up with the standard Coordinate object so we didn't have to do conversion back and forth. There might be too much work to double-check the existing assumptions (and the memory hit might be too large) to make this happen, though.

zerebubuth:
>  Is there a default config that's getting used somewhere?

Yup! There's a [default config function](https://github.com/tilezen/tilequeue/blob/c4a8d9365e2d1e0ff7f8ae1150a734d25c18be2c/tilequeue/config.py#L131), and the result of that is [merged into the config or replaces an empty config](https://github.com/tilezen/tilequeue/blob/c4a8d9365e2d1e0ff7f8ae1150a734d25c18be2c/tilequeue/config.py#L233).

That threw me, the first time I ran into it. Any way we can make it more obvious what's going on?

rmarianski:
The history with the configuration is that we used to, a long time ago in a galaxy far far away, support command line switches for each of the configuration values. The idea was that you could pass in a flag for a command, and have that trump configuration values, which would trump the default settings. The "merge" idea came from combining the input from all 3 of these sources, and providing a single point of access for the rest of the code, which is what the "Configuration" object is.

Since then, we don't have command line switches for anything except where the config file is, which we could manage more directly. We could gut out most of what's there, it's just that there wasn't much motivation to do so imho. If it's confusing to maintain though, let's go ahead and do it.

iandees:
Thanks for the pointers. It's likely ok as-is but I was going straight to the `Configuration` constructor and not seeing the call to `default_yml_config()`.

Since this "default config" can't actually run any of the tilequeue commands (so it's more of "null object" pattern than a default), does it make more sense to adjust the `Configuration` constructor so that it is resilient to having missing fields in the yaml?

rmarianski:
> does it make more sense to adjust the Configuration constructor so that it is resilient to having missing fields in the yaml?

Sure, although what can we do, if not provide default values? If it's clearer to just inline those defaults rather than have a separate section for it, that sounds fine to me.

iandees:
> If it's clearer to just inline those defaults rather than have a separate section for it, that sounds fine to me

Yea, maybe that's what I was describing. Regardless, I don't think it needs to happen in this PR.

config.yaml.sample
```diff
@@ -174,6 +174,18 @@ metatile:
...
+toi:
+  # We support storing the TOI in S3 or as a file
+  type: <file or s3>
+  file:
+    # The name of the file to store the TOI
+    name: toi.txt.gz
+  s3:
+    # The bucket and key for the s3 object that will contain the tiles of interest
+    bucket: test_bucket
+    key: toi/toi.txt.gz
```
rmarianski: :+1: thanks!

I like that you namespaced the configuration based on type. I wish I had done it that way for the other configuration sections instead of flattening it out, feels cleaner.

tilequeue/command.py
```diff
@@ -1423,77 +1410,59 @@ def tilequeue_dump_tiles_of_interest(cfg, peripherals):
...
 
-def tilequeue_load_tiles_of_interest(cfg, peripherals):
-    """
-    Given a newline-delimited file containing tile coordinates in
-    `zoom/column/row` format, load those tiles into a new Redis SET
-    alongside the existing one, then RENAME the new to the old.
-    """
-    logger = make_logger(cfg, 'load_tiles_of_interest')
-    logger.info('Loading tiles of interest')
+def tilequeue_dump_tiles_of_interest_from_redis(cfg, peripherals):
```
nvkelso: So this is here as a 1 time convienance for those migrating from earlier configurations, but we don't expect it to run more than once?

iandees: Right. I'll probably put together a PR after this one to remove this bit of code altogether.

nvkelso: Please add a comment explaining this bit (like you've done for tilequeue_load_tiles_of_interest below).

nvkelso: Okay, please file an issue so we don't loose track of it.

iandees: Fixed in d04ca1d.

PR #196, opened 2017-05-01
Remove Redis TOI dump command

iandees:
After we migrate the tiles of interest set from Redis in #193, we should merge this PR so that the command that dumps from Redis is removed.


nvkelso:
Since this removes functionality, it'll need a VERSION bump.

iandees:
I don't think this PR removes functionality or warrants a major version bump. This removes a temporary helper command to migrate us away from a Redis-based TOI and towards S3 or file-based TOI. The existing load- and dump- tiles of interest commands still exist.

I suppose removing support for Redis TOI in general could warrant a major rev.

nvkelso:
@iandees We've done our prod migration, can this PR be merged now?

iandees:
Yep!

PR #199, opened 2017-05-04
Support reading immortal tiles from s3

rmarianski:


tilequeue/command.py
```diff
@@ -1074,6 +1074,19 @@ def tilequeue_prune_tiles_of_interest(cfg, peripherals):
...
+            from boto.s3.bucket import Bucket
+            s3_conn = connect_s3()
+            bucket = Bucket(s3_conn, info['s3']['bucket'])
+            key = bucket.get_key(info['s3']['key'])
+            raw_coord_data = key.get_contents_as_string()
+            for line in raw_coord_data.splitlines():
+                coord = deserialize_coord(line.strip())
+                if coord:
+                    coord_int = coord_marshall_int(coord)
+                    uplifted_coord_int = coord_int_zoom_up(coord_int)
```
rmarianski: @iandees the coordinates that are currently in the file are what get requested, so they need to get uplifted. Do you think it makes more sense to have that happen on the way into the file in the first place? Or do you think it's better to handle it here when reading them as it stands now? It's weird that it's the only path that reads tiles and does the uplift.

tilequeue/command.py
```diff
@@ -1074,6 +1074,19 @@ def tilequeue_prune_tiles_of_interest(cfg, peripherals):
                     coord_marshall_int(deserialize_coord(l.strip()))
                     for l in f
                 )
+        elif 's3' in info:
+            from boto import connect_s3
+            from boto.s3.bucket import Bucket
+            s3_conn = connect_s3()
+            bucket = Bucket(s3_conn, info['s3']['bucket'])
+            key = bucket.get_key(info['s3']['key'])
+            raw_coord_data = key.get_contents_as_string()
```
iandees: I'm pretty sure you can iterate over `key` as a file-like so you don't have to do the `get_contents_as_string()` call.

tilequeue/command.py
iandees: Hmm, it might make sense to do the uplifting before it gets here. I agree that it's weird that this is the only path to do it. Maybe we can document somewhere that these "immortal tile" files/inputs are all assumed to be uplifted.

rmarianski:
> Hmm, it might make sense to do the uplifting before it gets here. I agree that it's weird that this is the only path to do it.

Stopped uplifting in 1e6abc5.

> Maybe we can document somewhere that these "immortal tile" files/inputs are all assumed to be uplifted.

Where does this make sense? Sample config file?

tilequeue/command.py
rmarianski: Good idea, but when I tried it locally, it seemed to iterate over "chunks" rather than lines. Is there a way to tell it to iterate over lines? Or is it better to just stick with the way it's done currently? I'm confident that this list will fit into memory :)

iandees: Eh, yea, let's do it this way :).

iandees:
> Where does this make sense? Sample config file?

Hm, yea. Sample config probably makes sense. Either that or in the code for the command.

PR #214, opened 2017-06-08
Reduce queue buffer size

rmarianski:


tilequeue/command.py
```diff
@@ -577,7 +577,8 @@ def tilequeue_process(cfg, peripherals):
     assert n_simultaneous_query_sets > 0
     # reduce queue size when we're rendering metatiles to try and avoid the
     # geometry waiting to be processed from taking up all the RAM!
-    default_queue_buffer_size = max(1, 128 >> (2 * (cfg.metatile_size or 0)))
+    default_queue_buffer_size = max(
+        1, 128 >> ((2 * (cfg.metatile_size or 0))) + 1)
```
rmarianski: This works for what I wanted to do (reduce it by a factor of 2), but probably not the best way to calculate it. Is there a better way to spell it?

iandees: Are you using bit shifting to represent a non-linear relationship? It'd be easier to read if it used Python's standard math syntax I think.

rmarianski: I'm not sure what the relationship should be. It should probably be related to metatile size, but imho if we wanted to provide a one size fits all formula we might need to take the available memory into consideration. Alternatively, we can just move this to configuration and have the existing calculation be the default. Or maybe not have any code changes at all here, because the downstream queue sizes use this just as a default, and already support being configured individually.

zerebubuth: I think the original idea was that the data size is probably proportional to the query area, which should be proportional to the `metatile_size` squared, and therefore the queue length should be inversely proportional.

However, that seems to have been translated effectively into `128 / (2 ** (metatile_size * 2))`, which gives answers which are close enough. Rather than adding a `+1`, it could be written as `64 >> (metatile_size * 2)`, but might it be easier to read as follows?

```python
size_sqr = (cfg.metatile_size or 1)**2
default_queue_buffer_size = max(1, 16 / size_sqr)
```

rmarianski: Sounds great to me. I used that exactly in ba670c8.

PR #221, opened 2017-06-29
Update output props processing for normalized sql

rmarianski:


rmarianski:
Assuming this is ok, because it's just plumbing the decisions that were made in the [vector-datasource pr](https://github.com/tilezen/vector-datasource/pull/1290).

PR #227, opened 2017-07-28
Makes queries per-table

rmarianski:
Connects to https://github.com/mapzen/tile-tasks/issues/267

Links up with https://github.com/tilezen/tileserver/pull/110 and https://github.com/tilezen/vector-datasource/issues/1352.

### DIFF
--- a/tilequeue/process.py
+++ b/tilequeue/process.py
@@ -475,7 +475,11 @@ def convert_source_data_to_feature_layers(rows, layer_data, bounds, zoom):
         boundaries_geometry = row.pop('__boundaries_geometry__', None)
         assert geometry or boundaries_geometry
 
-        common_props = row.pop('__properties__', None) or {}
+        common_props = row.pop('__properties__', None)
+        if common_props is None:
+            # if __properties__ exists but is null in the query, we
+            # want to normalize that to an empty dict too
+            common_props = {}
 
         row_props_by_layer = dict(
             boundaries=row.pop('__boundaries_properties__', None),
--- a/tilequeue/process.py
+++ b/tilequeue/process.py
@@ -445,7 +445,7 @@ def process_coord(coord, nominal_zoom, feature_layers, post_process_data,
     return all_formatted_tiles, extra_data
 
 
-def convert_source_data_to_feature_layers(rows, layer_data, bounds):
+def convert_source_data_to_feature_layers(rows, layer_data, bounds, zoom):
     # TODO we might want to fold in the other processing into this
     # step at some point. This will prevent us from having to iterate
     # through all the features again.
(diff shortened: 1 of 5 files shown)


tilequeue/process.py
```diff
@@ -443,3 +443,104 @@ def process_coord(coord, nominal_zoom, feature_layers, post_process_data,
         unpadded_bounds, cut_coords, buffer_cfg, extra_data, scale)
 
     return all_formatted_tiles, extra_data
+
+
+def convert_source_data_to_feature_layers(rows, layer_data, bounds, zoom):
+    # TODO we might want to fold in the other processing into this
```
nvkelso: Is this a note for the glorious future or for the current sprint?

tilequeue/process.py
```diff
@@ -443,3 +443,104 @@ def process_coord(coord, nominal_zoom, feature_layers, post_process_data,
...
+                if label_geometry:
+                    query_props['__label__'] = label_geometry
+
+                features_by_layer[layer].append(query_props)
+
+    feature_layers = []
+    for layer_datum in layer_data:
+        layer = layer_datum['name']
+        features = features_by_layer[layer]
+        # TODO padded bounds
```
nvkelso: To support the MVT Buffered format or something additional?

tilequeue/process.py
rmarianski: This would have been a performance improvement, if we find that when we run it on dev it's too slow. If it's unnecessary, it might be better to hold off on this since it might change with phase 3 anyway.

tilequeue/process.py
rmarianski: Yes, exactly.

tilequeue/process.py
```diff
@@ -443,3 +443,104 @@ def process_coord(coord, nominal_zoom, feature_layers, post_process_data,
...
+    for row in rows:
+
+        fid = row.pop('__id__')
+
+        geometry = row.pop('__geometry__', None)
+        label_geometry = row.pop('__label__', None)
+        boundaries_geometry = row.pop('__boundaries_geometry__', None)
+        assert geometry or boundaries_geometry
+
+        common_props = row.pop('__properties__', None) or {}
```
zerebubuth: Is this the same thing as `row.pop('__properties__', {})`?

zerebubuth: Oh, I see... the way it is at the moment handles the case where `row['__properties__']` exists, but is `None`. I feel like that's slightly non-obvious.

rmarianski: I just confused myself too, so I expanded it out and added a comment in da19aa0.

PR #229, opened 2017-08-02
Fixture-based data source

zerebubuth:
This adds a fixture-based data source, which returns data in a similar format to the data coming from PostgreSQL, so that it can be processed in the same way by the rest of the system. The idea is to use it for tests before widening the implementation to include RAWR tiles.

Starts to lay the foundation for tilezen/vector-datasource#923.

### DIFF
--- a/tilequeue/query/fixture.py
+++ b/tilequeue/query/fixture.py
@@ -1,0 +1,56 @@
+from collections import namedtuple
+from shapely.geometry import box
+
+
+LayerInfo = namedtuple('LayerInfo', 'min_zoom_fn props_fn')
+
+
+class DataFetcher(object):
+
+    def __init__(self, layers, rows):
+        """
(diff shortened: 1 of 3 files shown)


tilequeue/query/fixture.py
```diff
@@ -0,0 +1,56 @@
...
+        for (fid, shape, props) in self.rows:
+            # reject any feature which doesn't intersect the given bounds
+            if bbox.disjoint(shape):
+                continue
+
+            # place for assembing the read row as if from postgres
+            read_row = {}
+
+            for layer_name, info in self.layers.items():
+                # TODO: what should meta be?
```
nvkelso: Is there an issue that tracks this TODO?

zerebubuth: Added as #230. This issue is somewhat nebulous at this point, but I suspect that it'll become clearer as we try to implement more parts of the system.

PR #231, opened 2017-08-03
Remove hard coded list of layers in conversion

rmarianski:


rmarianski:
Yea,  originally I don't think I was taking in the `layer_data` in that function, but eventually I needed it and never updated the hard coded list.

PR #232, opened 2017-08-04
Add a tile status command

zerebubuth:
This command can be given a list of tiles and it will print out its in-flight, TOI and storage status. This can be helpful while debugging "stuck" tiles, although in the time it took me to write the tool, the issue [was probably self-fixed by the gardener](https://github.com/tilezen/vector-datasource/issues/1363#issuecomment-320227790).

I had to do some work-around-y stuff to pass the positional `coords` arguments to the `command_tile_status` function, so I'd very much appreciate any recommendations about how that might be done more neatly or cleanly.

### DIFF
--- a/tilequeue/command.py
+++ b/tilequeue/command.py
@@ -1547,12 +1547,12 @@ def tilequeue_tile_status(cfg, peripherals, args):
         # input checking! make sure that the coordinate is okay to use in
         # the rest of the code.
         if not coord:
-            logger.warning('Could not deserialize %r as coordinate' \
+            logger.warning('Could not deserialize %r as coordinate'
                            % (coord_str,))
             continue
 
         if not coord_is_valid(coord):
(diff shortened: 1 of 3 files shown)


tilequeue/command.py
```diff
@@ -1528,6 +1519,65 @@ def tilequeue_load_tiles_of_interest(cfg, peripherals):
...
+
+        if not coord_is_valid(coord):
+            logger.warning('Coordinate is not valid: %r (parsed from %r)'
+                           % (coord, coord_str))
+            continue
+
+        # now we think we probably have a valid coordinate. go look up
+        # whether it exists in various places.
+
+        logger.info("=== %s ===" % (coord_str,))
```
iandees: Logger will do string formatting for you. If you want you can use

```
logger.info("=== %s ===", coord_str)
```

zerebubuth: :+1:

Thanks! I didn't know that. It makes the code much more readable (bc007b3).

PR #234, opened 2017-08-07
Support lists of templates per source

rmarianski:
Connects to https://github.com/tilezen/vector-datasource/pull/1365

### DIFF
--- a/tilequeue/query/postgres.py
+++ b/tilequeue/query/postgres.py
@@ -239,7 +239,7 @@ def parse_source_data(queries_cfg):
         for template_data in templates:
             template = template_data['template']
             zoom_start = int(template_data.get('zoom_start', 0))
-            zoom_stop = int(template_data.get('zoom_stop', 25))
+            zoom_stop = int(template_data.get('zoom_stop', 21))
             template_spec = TemplateSpec(template, zoom_start, zoom_stop)
             template_specs.append(template_spec)
         source = DataSource(source_name, template_specs)
--- a/tilequeue/query/postgres.py
+++ b/tilequeue/query/postgres.py
@@ -7,11 +7,8 @@ from tilequeue.transform import calculate_padded_bounds
 import sys
 
 
-DataSource = namedtuple('DataSource', 'name template, start_zoom')
-
-
-def make_source(name, template, start_zoom):
-    return DataSource(name, template, start_zoom)
+TemplateSpec = namedtuple('TemplateSpec', 'template zoom_start zoom_stop')
+DataSource = namedtuple('DataSource', 'name template_specs')
 
 
 class TemplateFinder(object):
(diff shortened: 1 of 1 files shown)


tilequeue/query/postgres.py
```diff
@@ -231,10 +234,15 @@ def make_queries_generator(sources, template_path, reload_templates):
...
-    for source_name, source_data in sources_cfg.items():
-        template = source_data['template']
-        start_zoom = int(source_data.get('start_zoom', 0))
-        source = make_source(source_name, template, start_zoom)
+    for source_name, templates in sources_cfg.items():
+        template_specs = []
+        for template_data in templates:
+            template = template_data['template']
+            zoom_start = int(template_data.get('zoom_start', 0))
+            zoom_stop = int(template_data.get('zoom_stop', 25))
```
nvkelso: Do we say 21 elsewhere?

tilequeue/query/postgres.py
```diff
@@ -67,9 +64,15 @@ def __init__(self, sources, query_generator):
     def __call__(self, zoom, bounds):
         queries = []
         for source in self.sources:
-            if source.start_zoom <= zoom:
-                query = self.query_generator(source.template, bounds, zoom)
-                queries.append(query)
+            template_queries = []
+            for template_spec in source.template_specs:
+                if template_spec.zoom_start <= zoom < template_spec.zoom_stop:
```
zerebubuth: Should this be `zoom_start <= zoom <= zoom_stop`? In other words, is the start to end range of zooms inclusive of the stop or exclusive?

tilequeue/query/postgres.py
rmarianski: This just needs to be a larger number, but you're right that it's better to be consistent. I updated that in a90837d.

tilequeue/query/postgres.py
rmarianski: Exclusive. I put in a comment in the [queries.yaml in the vector-datasource pr](https://github.com/tilezen/vector-datasource/pull/1365/files#diff-fc49e5c256610bc0b32a7de5a0520835R20), but that doesn't help reading the code. I added in comments in 0e9e0db.

Would you prefer that it just be renamed to something more verbose but clear like `zoom_stop_exclusive`? It's nice that start/stop feel symmetric, but having the extra characters for clarity seems worthwhile, and don't bother me :)

zerebubuth: I think it should be inclusive. IIRC, we're inclusive about zoom ranges in the post-processing functions, [e.g](https://github.com/tilezen/vector-datasource/blob/4eb1fbd1cab6194713761470b2fdb57eb43d3c01/vectordatasource/transform.py#L2072-L2076).

On second thoughts, I don't mind whether it's inclusive or exclusive, as long as we're consistent. I think the one-past-the-end counting is easier for us to reason about internally, but perhaps the inclusive ranges are more intuitive for non-developers?

rmarianski: My personal opinion is that exclusive just reads better here, but that could just be me being used to that coming from a programming context. The ranges in the sql were already exclusive, and imho it's clearer to follow what's going on.

For example, this is typical:

```
zoom_start: 0
zoom_stop: 2

zoom_start: 2
zoom_stop: 5

zoom_start: 5
zoom_stop: 8
```

Spelled inclusively:

```
zoom_start: 0
zoom_stop: 1

zoom_start: 2
zoom_stop: 4

zoom_start: 5
zoom_stop: 7
```

About consistency with the other configuration in the transforms, maybe we should just update it all to be exclusive across the board? Or, maybe we just strictly adhere to a naming convention to clearly delineate which is used. Initially I had been using the term `zoom_until` to mean inclusive, but it looks like that's fallen out of fashion because I don't see it there any more.

rmarianski: I should have noted that I'm happy to change it to be inclusive if that's what we prefer overall. @nvkelso what do you think?

zerebubuth: I've added #235 for deciding which makes more sense and then making all the code consistent. It seems like something that might need to be discussed more widely, and I don't want to hold this PR up.

PR #236, opened 2017-08-08
Move tilequeue.postgresql -> tilequeue.query.pool

rmarianski:
The db connection pool logic seemed like a good candidate for moving
under the tilequeue.query location.


tilequeue/query/__init__.py
```diff
@@ -1,4 +1,9 @@
-from tilequeue.query import postgres, fixture
+from tilequeue.query.fixture import make_data_fetcher
+from tilequeue.query.pool import DBConnectionPool
+from tilequeue.query.postgres import make_db_data_fetcher
 
-make_db_data_fetcher = postgres.make_db_data_fetcher
-make_fixture_data_fetcher = fixture.make_data_fetcher
+__all__ = [
+    'DBConnectionPool',
+    'make_data_fetcher',
```
zerebubuth: I was using the assignment to rename `fixture.make_data_fetcher` to `make_fixture_data_fetcher`. Is that possible with the `__all__` re-export, or should we just rename the function in the `fixture.py` file?

rmarianski: Doh, sorry I missed that. Is the rename important? The tests pass for me locally without it.

If we do want the rename, I don't think `__all__` can do it and we should just set a variable at module scope like you had it before for that particular import.

rmarianski: Reading it more closely, I can see why you wanted to do that. Updated in 2a71d68.

PR #237, opened 2017-08-08
Correct function used for parser

rmarianski:
Previously, the command fn bound for a parser would always be the last
one. This creates a new function instead of each binding to the same
one.


rmarianski:
> Ooops! So it only worked for me because the command/function I was using was the last one, which I just added? D'oh!

Bad luck :) I didn't catch it in the review either, just noticed it now when I happened to try and run a different command.

PR #239, opened 2017-08-11
Add tilequeue command to list stuck tiles

rmarianski:
Related to root cause analysis of https://github.com/tilezen/vector-datasource/issues/1363 and https://github.com/tilezen/vector-datasource/issues/1320. This new script will be used to list and then delete stuck tiles.

Followup will be filed relating to TOI Gardener config for root cause fix.


rmarianski:
> Might it be worth moving the tile listing functionality into the store classes? For example, adding a method list_tiles(layer, format) which returns a generator.

Yea, it's a good idea, but I was just being more reserved about adding functions to that api. Listing the tiles felt more like a one-off task rather than something that should be supported by all implementations, but happy to add it to all of them if you think that ends up being cleaner.

zerebubuth:
> I was just being more reserved about adding functions to that api. Listing the tiles felt more like a one-off task rather than something that should be supported by all implementations

That's a very good point. I guess I felt that we should try and keep the S3-specific stuff in the S3 storage implementation. I think it's fine to have a stub implementation in the file storage which just does `raise UnimplementedError()`. The implementation over `walk` is probably pretty trivial, in case anyone wants to add that in the future.

rmarianski:
I moved the implementation over to the s3 store, and added implementations for the other stores as well.

PR #243, opened 2017-09-18
Modify Unit of work on queue

rmarianski:
WIP still, but putting it up if you want to take a look to see if I'm taking the right approach.

I waffled a bit, but ended up settling on modifying the queue implementations to work with raw payloads, and not coordinates. This means that we'll have more concepts and moving parts, but imho is a better design because it means we can mix and match them more readily which will allow us to be more flexible, and it keeps each component simple.
As part of this, I also teased apart the in flight management, so the only responsibility for queue implementations now is to translate calls to their underlying backing queue.

The additional concepts:

* message marshaller - translates a payload from an individual queue message into multiple coordinates
* message tracker - coordinates between work on the individual tiles and the upstream queue message they correspond to. This will manage the mapping between message handles that tile queues understand, and coordinate handles that the processing pipelines work with.
*  message grouper - takes in a sequence of coordinates, and groups them into cohesive chunks for sending to the queue. In this case, it'll group z10+ tiles into a single chunk based on common z10 parents.
* inflight manager - takes in a sequence of coordinates, and filters out the coordinates that are in the in flight list.

I've also added a queue "writer", which threads all these together. I've verified that this works with test sqs queues, but I haven't updated all the existing code to use this yet. This would be the next step, and it would be a matter of just ensuring that we have configuration to dial this in, and use this everywhere we have the threaded enqueuer currently, eg seeding and intersection code paths.

There is a complication that came up thus far, which is that fundamentally the multisqs queue implementation needs to know the zoom level that the payload corresponds to in order to dispatch to the appropriate queue. I handled this by having the enqueue methods take in metadata to specify this at the moment. This is working, but it might be better to instead remove the multisqs implementation and just formalize this bit of logic into its own component and add it to the writer. The mapping logic already gets constructed outside the multisqs implementation (it just gets a lookup object), so this isn't such a big step. Now that I think about it more, I'm leaning towards moving in this direction.

I also plan on adding more tests throughout, reviewing the existing ones more thoroughly (I only just got them to pass), and documenting these ideas more in the code itself, but I think it's prudent to ask for a review now before spending more time on that.


tilequeue/queue/grouper.py
```diff
@@ -0,0 +1,46 @@
...
+    Take a sequence of coordinates, and emit a sequence of CoordGroup
+    instances. These contains coords grouped by their parent zoom, and
+    metadata that contains the parent zoom used.
+    """
+
+    def __init__(self, start_zoom_to_group):
+        self.start_zoom_to_group = start_zoom_to_group
+
+    def __call__(self, coords):
+        grouped = {}
```
zerebubuth: I found this a little hard to understand. Would it be more readable as:

```python
grouped = defaultdict(list)
for coord in coords:
    if coord.zoom >= self.start_zoom_to_group:
        # assuming zoomTo().container() is a no-op if the zooms are the same?
        coord_at_group_zoom = coord.zoomTo(self.start_zoom_to_group).container()
        group_key = coord_marshall_int(coord_at_group_zoom)
        grouped[group_key].append(coord)
    else:
         ...
```

?

tilequeue/queue/inflight.py
```diff
@@ -0,0 +1,29 @@
...
+        self.inflight_key = inflight_key
+        self.chunk_size = chunk_size
+
+    def filter(self, coords):
+        for coord in coords:
+            coord_int = coord_marshall_int(coord)
+            if not self.redis_client.sismember(self.inflight_key, coord_int):
+                yield coord
+
+    def mark_inflight(self, coords):
```
zerebubuth: Does there need to be a corresponding `unmark_inflight` (`mark_landed`?) function called from somewhere to call the `redis_client.srem()` to prevent us getting "stuck" tiles?

tilequeue/queue/writer.py
```diff
@@ -0,0 +1,47 @@
...
+        # used for metadata
+        zooms = []
+        for payload, coords, metadata in queue_send_data_chunk:
+            payloads.append(payload)
+            all_coords.extend(coords)
+            zoom = metadata.get('zoom')
+            assert zoom is not None
+            zooms.append(zoom)
+
+        metadata = dict(zooms=zooms)
```
zerebubuth: I'm confused about what's going on here. The coords have been grouped, each one getting a single `zoom` in the metadata, then we batch the groups, adding the list of zooms as metadata? It looks like they just get `izip`ped back together in the SQS queue writer.

tilequeue/queue/grouper.py
rmarianski: Sure, happy to move to using a `defaultdict` here instead.

tilequeue/queue/inflight.py
rmarianski: Ah right, I'm missing that bit at the moment. I'll go ahead and add that in.

nvkelso:
Rob and I chatted IRL about this but I'd like to look at 4 queues:

- Hotfix path
- zoom 0-9 low-zoom
- zoom 10+ TOI
- zoom 10+ nonTOI

tilequeue/queue/writer.py
rmarianski: Yea, it's a bit confusing/convoluted at the moment. The initial grouping is by zoom level to serialize the payload, and then the multisqs is grouping again based on the queue that it will send to, which is also based on zoom. I think this also points to removing the multisqs queue that I was hinting at above, in favor of having all the grouping happen in one place. I think I'll just move in that direction and I'm hoping that the code will become simpler here.

tilequeue/queue/grouper.py
rmarianski: Updated in 4d2f306.

tilequeue/queue/inflight.py
rmarianski: Initial bit of code to unmark coords from in flight list in aee7802.

nvkelso:
From today's Tile sync meeting:

- We should also allow stratification based on zoom inside the message packs (more like how we do priority queues now). That’d be ~3 statra groups resulting in 3 S3 reads instead of "1000" db requests.

rmarianski:
@zerebubuth with the latest changes, this pr should be more or less ready to go. Mind giving it another review to see what you think?

PR #244, opened 2017-10-03
Add RAWR data fetcher

zerebubuth:
This adds a RAWR compatible data fetcher, although it isn't hooked up to actually download tiles at the moment. The only use of it is from the test suite. The next step will be to hook it into the rendering pipeline and allow configuring it.

Several issues with the current approach (shared with the fixture data fetcher):

1. There's a lot of hard-coded stuff that relates to "style" decisions that seem like they'd be better in `vector-datasource`. Hopefully once the code is working, we can pull those out and put them wherever would be more appropriate.
2. The staging in the indexing is quite delicate, to make sure that data is read only once. This might make it brittle, but it's hard to tell when it's just running through the test suite.


tilequeue/query/rawr.py
```diff
@@ -0,0 +1,578 @@
...
+class TilePyramid(namedtuple('TilePyramid', 'z x y max_z')):
+    """
+    Represents a "tile pyramid" of all tiles which are geographically
+    contained within the tile `z/x/y` up to a maximum zoom of `max_z`. This is
+    the set of tiles corresponding to one RAWR tile.
+    """
+
+    def tile(self):
+        from raw_tiles.tile import Tile
+        return Tile(self.z, self.x, self.y)
```
rmarianski: I wonder if we should at some point just create our own repo to hold these kinds of objects and related utility functions?

tilequeue/query/rawr.py
```diff
@@ -0,0 +1,578 @@
...
+    point = 1
+    line = 2
+    polygon = 3
+
+    _LOOKUP = ['point', 'line', 'polygon']
+
+    @staticmethod
+    def lookup(typ):
+        "turn the enum into a string"
+        return ShapeType._LOOKUP[typ-1]
```
rmarianski: What do you think of using the [enum package](https://pypi.python.org/pypi/enum34) here?

tilequeue/query/rawr.py
zerebubuth: Follow-up in #246.

tilequeue/query/rawr.py
zerebubuth: Good point. At first, it didn't seem necessary to bring that in for something which, at least initially, was very simple. After adding the `lookup` function, it's starting to look like it would be cleaner with the `enum` package. Fixed in #247.

PR #248, opened 2017-10-06
Add rawr enqueue and process commands

rmarianski:
The enqueueing and rawr tile generation is working for me locally. There's more that needs to be added to it, like error handling and logging, but I wanted to put this up for review first just to make sure the current implementation was adequate before spending effort on that. I was thinking that those kinds of improvements can be done in a subsequent pr.

### DIFF
--- a/tilequeue/rawr.py
+++ b/tilequeue/rawr.py
@@ -5,6 +5,7 @@ from ModestMaps.Core import Coordinate
 from raw_tiles.tile import Tile
 from tilequeue.command import explode_and_intersect
 from tilequeue.queue.message import MessageHandle
+from tilequeue.store import calc_hash
 from tilequeue.tile import coord_marshall_int
 from tilequeue.tile import coord_unmarshall_int
 from tilequeue.toi import load_set_from_gzipped_fp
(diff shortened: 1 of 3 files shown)


rmarianski:

tilequeue/rawr.py
```diff
@@ -0,0 +1,254 @@
...
+        else:
+            assert parent == coord_parent
+    assert parent is not None, 'No coords?'
+    return parent
+
+
+def convert_coord_object(coord):
+    """Convert ModestMaps.Core.Coordinate -> raw_tiles.tile.Tile"""
+    assert isinstance(coord, Coordinate)
+    return Tile(coord.zoom, coord.column, coord.row)
```
zerebubuth: `Tile` was intended to take `int` coordinates. Would it be worth either asserting that here, and/or calling `.container()` on the `Coordinate`?

tilequeue/rawr.py
```diff
@@ -0,0 +1,254 @@
...
+            queue_writer, rawr_toi_intersector):
+        self.rawr_queue = rawr_queue
+        self.msg_marshaller = msg_marshaller
+        self.group_by_zoom = group_by_zoom
+        self.rawr_gen = rawr_gen
+        self.queue_writer = queue_writer
+        self.rawr_toi_intersector = rawr_toi_intersector
+
+    def __call__(self):
+        # TODO - should probably provide a way to stop iteration
```
zerebubuth: I agree :wink:

tilequeue/rawr.py
rmarianski: Good call, updated in 701ab96.

tilequeue/rawr.py
```diff
@@ -0,0 +1,254 @@
...
+    buf = StringIO()
+    with zipfile.ZipFile(buf, mode='w') as z:
+        for fmt_data in rawr_tile.all_formatted_data:
+            zip_info = zipfile.ZipInfo(fmt_data.name, date_time)
+            z.writestr(zip_info, fmt_data.data, zipfile.ZIP_DEFLATED)
+    return buf.getvalue()
+
+
+def make_s3_path(tile, prefix):
+    return '%s/%d/%d/%d.zip' % (prefix, tile.z, tile.x, tile.y)
```
zerebubuth: Should we be hash-prefixing the tiles? I guess there's only a million of them, but it would be theoretically nice to be able to write all million in parallel (which means having at least 3 hex digits of hash - 5,000 partitions).

rmarianski: Probably, good catch!

I added that in cb6ad62. I also had it take a suffix for the extension, which most likely doesn't make a difference practically, but it's how everything else is stored too.

tilequeue/rawr.py
rmarianski: Want this added as part of this pr? Would a simple boolean, eg `stopped` or `done`, on the instance suffice?

zerebubuth: Actually, I take that back.

In `tilequeue_process`, we never stop - just trap a bunch of signals to shut down the threads. In this case, since there aren't any threads, we don't need to do that and a `^C` will suffice to shut it down cleanly enough.

Perhaps we don't need a way to stop iteration? Would we run this as a separate thread / process under the main `tilequeue_process`? Or as a separate service managed by runit?

rmarianski: > Perhaps we don't need a way to stop iteration? Would we run this as a separate thread / process under the main tilequeue_process? Or as a separate service managed by runit?

Separate service managed by runit. So yea, we technically don't need to stop iteration, because it's single threaded. Maybe we add some kind of stop mechanism if we add threads?

zerebubuth: :+1: Sounds good to me.

PR #256, opened 2017-10-17
Add json logging to tilequeue processing

rmarianski:

A new logging abstraction was created, and the processing worker classes now all use it for logging.

I expect this to conflict with https://github.com/tilezen/tilequeue/pull/255. I'm happy to resolve any conflicts once that gets merged, since the changes I've made are all just logging based and will not have changed any of the logic.

### DIFF
--- a/tilequeue/log.py
+++ b/tilequeue/log.py
@@ -31,10 +31,7 @@ def log_level_name(log_level):
 
 
 def log_category_name(log_category):
-    if log_category == LogCategory.QUEUE_SIZES:
-        return 'queue-sizes'
-    else:
-        return log_category.name.lower()
+    return log_category.name.lower()
 
 
 class JsonTileProcessingLogger(object):
--- a/tilequeue/log.py
+++ b/tilequeue/log.py
@@ -1,0 +1,111 @@
+from enum import Enum
+import json
+import logging
+import sys
+
+
+def make_coord_dict(coord):
+    """helper function to make a dict from a coordinate for logging"""
+    return dict(
+        z=coord.zoom,
+        x=coord.column,
+        y=coord.row,
+    )
+
+
+class LogLevel(Enum):
+    DEBUG = logging.DEBUG
+    INFO = logging.INFO
+    WARNING = logging.WARNING
+    ERROR = logging.ERROR
+
+
+class LogCategory(Enum):
+    PROCESS = 1
+    LIFECYCLE = 2
+    QUEUE_SIZES = 3
+
+
+def log_level_name(log_level):
+    return log_level.name.lower()
+
+
+def log_category_name(log_category):
+    if log_category == LogCategory.QUEUE_SIZES:
+        return 'queue-sizes'
+    else:
+        return log_category.name.lower()
+
+
+class JsonTileProcessingLogger(object):
+
+    def __init__(self, logger):
+        self.logger = logger
+
+    def log(self, log_level, log_category, msg, exception,
+            formatted_stacktrace, coord):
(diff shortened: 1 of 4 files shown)


nvkelso:
> I expect this to conflict with #255. I'm happy to resolve any conflicts once that gets merged, since the changes I've made are all just logging based and will not have changed any of the logic.

👍

tilequeue/log.py
```diff
@@ -0,0 +1,111 @@
...
+    QUEUE_SIZES = 3
+
+
+def log_level_name(log_level):
+    return log_level.name.lower()
+
+
+def log_category_name(log_category):
+    if log_category == LogCategory.QUEUE_SIZES:
+        return 'queue-sizes'
```
nvkelso: we seem to use `_` everywhere else, is the `-` deliberate here?

(I just solved a related _ versus - bug in the basemap styles so sensitive to this...)

tilequeue/worker.py
```diff
@@ -484,45 +497,33 @@ def __call__(self):
                 if tile_timestamp_millis is not None:
                     tile_timestamp_seconds = tile_timestamp_millis / 1000.0
                     time_in_queue = now - tile_timestamp_seconds
+            timing['queue'] = time_in_queue
 
             layers = metadata['layers']
             size = layers['size']
-            size_as_str = repr(size)
+            # size_as_str = repr(size)
```
nvkelso: Should this be removed entirely?

tilequeue/log.py
rmarianski: Good call. I checked how we do the tapalcatl logging, and we also prefer `_` there too. Updated in 4310ce1.

tilequeue/worker.py
rmarianski: Yea, agree. Removed in ad9c83d.

tilequeue/log.py
```diff
@@ -0,0 +1,111 @@
...
+            logging_log_level = log_level.value
+        except:
+            print >>sys.stderr, ('ERROR: code error: invalid log level: %s' %
+                                 log_level)
+            log_level_str = log_level_name(LogLevel.ERROR)
+            logging_log_level = logging.ERROR
+        try:
+            log_category_str = log_category_name(log_category)
+        except:
+            print >>sys.stderr, ('ERROR: code error: invalid log category: %s'
```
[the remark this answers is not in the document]
rmarianski: Makes sense, updated in aa89b44.

tilequeue/worker.py
```diff
@@ -62,11 +64,10 @@ def __call__(self, coord, data):
                 if self.stop.is_set():
                     return True
 
-        except MemoryError:
+        except MemoryError as e:
```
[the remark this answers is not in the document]
rmarianski: To me, the way it is now seems more explicit and straightforward. I'm inclined to just leave it the way it is unless there are objections.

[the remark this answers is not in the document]
iandees: Yep, sounds good to me.

rmarianski:
Rebased the changes off of master.

tilequeue/worker.py
```diff
@@ -491,6 +502,15 @@ def __call__(self):
             coord = data['coord']
 
             start = time.time()
+
+            try:
+                self.inflight_mgr.unmark_inflight(coord)
```
zerebubuth: Nice catch! Moving this before the queue job acknowledgement will mean we're less likely to leave a tile marked as in-flight when it's not.

PR #258, opened 2017-10-18
Add stats handling

rmarianski:

Updates processing metrics to send over in code now, instead of indirectly through logs via logstash.

Added new metrics for rawr tile enqueueing and generation.

### DIFF
--- a/tilequeue/rawr.py
+++ b/tilequeue/rawr.py
@@ -87,11 +87,13 @@ class SqsQueue(object):
 class RawrEnqueuer(object):
     """enqueue coords from expiry grouped by parent zoom"""
 
-    def __init__(self, rawr_queue, msg_marshaller, group_by_zoom, logger):
+    def __init__(self, rawr_queue, msg_marshaller, group_by_zoom, logger,
+                 stats_handler):
         self.rawr_queue = rawr_queue
         self.msg_marshaller = msg_marshaller
         self.group_by_zoom = group_by_zoom
         self.logger = logger
+        self.stats_handler = stats_handler
 
     def __call__(self, coords):
         grouped_by_zoom = defaultdict(list)
@@ -120,6 +122,8 @@ class RawrEnqueuer(object):
                 'Expiry processed: '
                 'coords(%d) payloads(%d) enqueue-calls(%d))' %
                 (n_coords, n_payloads, n_msgs_sent))
+
+        self.stats_handler(n_coords, n_payloads, n_msgs_sent)
 
 
 def common_parent(coords, parent_zoom):
@@ -334,8 +338,10 @@ def make_rawr_s3_path(tile, prefix, suffix):
     return path_with_hash
 
 
-def make_rawr_enqueuer(rawr_queue, msg_marshaller, group_by_zoom, logger):
-    return RawrEnqueuer(rawr_queue, msg_marshaller, group_by_zoom, logger)
+def make_rawr_enqueuer(rawr_queue, msg_marshaller, group_by_zoom, logger,
+                       stats_handler):
+    return RawrEnqueuer(rawr_queue, msg_marshaller, group_by_zoom, logger,
+                        stats_handler)
 
 
 class RawrS3Sink(object):
--- a/tilequeue/rawr.py
+++ b/tilequeue/rawr.py
@@ -16,6 +16,7 @@ from tilequeue.utils import grouper
 from tilequeue.utils import time_block
 from time import gmtime
 import json
+import sys
 import zipfile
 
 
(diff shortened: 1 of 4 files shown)


tilequeue/stats.py
```diff
@@ -0,0 +1,63 @@
...
+class RawrTileEnqueueStatsHandler(object):
+
+    def __init__(self, stats):
+        self.stats = stats
+
+    def __call__(self, n_coords, n_payloads, n_msgs_sent):
+        with self.stats.pipeline() as pipe:
+            pipe.gauge('rawr.enqueue.coords', n_coords)
+            pipe.gauge('rawr.enqueue.groups', n_payloads)
+            pipe.gauge('rawr.enqueue.calls', n_msgs_sent)
```
rmarianski: This name is a little opaque. It represents the number of function calls we make to the underlying sqs queue 
 client. Each call will have up to 10 payloads which correspond to sqs messages, and each payload can have multiple tiles in it, which are z10 groupings.

tilequeue/stats.py
```diff
@@ -0,0 +1,63 @@
...
+
+class RawrTileEnqueueStatsHandler(object):
+
+    def __init__(self, stats):
+        self.stats = stats
+
+    def __call__(self, n_coords, n_payloads, n_msgs_sent):
+        with self.stats.pipeline() as pipe:
+            pipe.gauge('rawr.enqueue.coords', n_coords)
+            pipe.gauge('rawr.enqueue.groups', n_payloads)
```
rmarianski: Each group corresponds to a payload or sqs message, and represents a set of tiles which share the same z10 parent. Is there a better name for it?

tilequeue/stats.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+class RawrTileEnqueueStatsHandler(object):
+
+    def __init__(self, stats):
+        self.stats = stats
+
+    def __call__(self, n_coords, n_payloads, n_msgs_sent):
+        with self.stats.pipeline() as pipe:
+            pipe.gauge('rawr.enqueue.coords', n_coords)
```
rmarianski: The number of coordinates that came from the expiry list.

tilequeue/rawr.py
```diff
@@ -256,26 +263,33 @@ def __call__(self):
...
 
                 with time_block(timing, 'queue-write'):
-                    self.queue_writer.enqueue_batch(coords_to_enqueue)
+                    n_enqueued, n_inflight = \
+                        self.queue_writer.enqueue_batch(coords_to_enqueue)
                 coords_enqueued = True
 
                 with time_block(timing, 'queue-done'):
                     self.rawr_queue.done(msg_handle)
+                msg_done = True
```
zerebubuth: I found the flags somewhat confusing, e.g:

```python
try:
  do_first_thing(...)
  first_thing_done = True
  do_second_thing(...)
  second_thing_done = True
  ...
except:
  if not first_thing_done:
    ...
  elif not second_thing_done:
    ...
```

Might it be clearer written out as individual try-catch blocks, perhaps wrapping the logging to avoid repetition?

```python
try:
  do_first_thing(...)
except Exception:
  self._log_exception("first thing")
  continue

try:
  do_second_thing(...)
except Exception:
  self._log_exception("second thing")
  continue

...
def _log_exception(self, msg):
  if self.logger:
    stacktrace = format_stacktrace_one_line()
    msg = 'Error: %s' % msg
    ...
```

Another possibility might be to have a single `try` block, but modify the exceptions with an internal `with` block, something like:

```python
try:
  with describe_error("first thing"):
    do_first_thing(...)
  with describe_error("second thing"):
    do_second_thing(...)

except Exception:
  if self.logger:
    self.logger.info(e.message())

class describe_error:
  def __init__(self, msg):
    self.msg = msg

  def __enter__(self):
    pass

  def __exit__(self, type, value, traceback):
    raise Exception(self.msg, parent=value, traceback=traceback)
```

What do you think?

tilequeue/stats.py
zerebubuth: This is the number of SQS API calls?

tilequeue/stats.py
zerebubuth: I'm not sure. I think we've probably overloaded all the words at this point :wink: Groups sounds fine to me.

tilequeue/stats.py
rmarianski: Yea, exactly.

tilequeue/rawr.py
rmarianski: Agreed that this was a little clunky. It started off with just one or two flags, but then grew into many more.

I considered the `with` block route, but I wasn't immediately sure whether the stacktrace would be preserved as expected by chaining the exceptions. We could instead just mutate some state that contained the error message, similarly to the timing block, but that seemed like it was a little bit awkward or heavy handed for what it was trying to do. I decided to just wrap each block in its own try/except like you mentioned first, and although it's repetitive, it's at least clear as to what's going on.

Updated in 1548e66.

PR #259, opened 2017-10-19
Add rawr-seed-toi command

rmarianski:

This command will effectively replace the other seed command for our purposes. Higher zoom level tiles will get enqueued onto the rawr queue that need to get grouped, and after the rawr tile gets generated, the children tiles of that rawr tile will get enqeueud for processing. Low zoom level tiles will get enqueued directly to the tilequeue processing queues.

If we wanted to split this apart a little bit more in the future, we'd need to have the consumers know that these tiles enqueued for processing are for rawr tile generation only, and not for subsequent tilequeue processing, as is the typical workflow after applying diffs.

### DIFF
--- a/logging.conf.sample
+++ b/logging.conf.sample
@@ -1,5 +1,5 @@
 [loggers]
-keys=root,process,seed,prune_tiles_of_interest,enqueue_tiles_of_interest,dump_tiles_of_interest,load_tiles_of_interest,wof_process_neighbourhoods,query,consume_tile_traffic,tile_status,stuck_tiles,delete_stuck_tiles,rawr_enqueue,rawr_process
+keys=root,process,seed,prune_tiles_of_interest,enqueue_tiles_of_interest,dump_tiles_of_interest,load_tiles_of_interest,wof_process_neighbourhoods,query,consume_tile_traffic,tile_status,stuck_tiles,delete_stuck_tiles,rawr_enqueue,rawr_process,rawr_seed
 
 [handlers]
 keys=consoleHandler,jsonConsoleHandler
@@ -93,6 +93,12 @@ propagate=0
 level=INFO
 handlers=consoleHandler
 qualName=rawr_process
+propagate=0
+
+[logger_rawr_seed]
+level=INFO
+handlers=consoleHandler
+qualName=rawr_seed
 propagate=0
 
 [handler_consoleHandler]
--- a/tilequeue/command.py
+++ b/tilequeue/command.py
@@ -1208,27 +1208,13 @@ def tilequeue_prune_tiles_of_interest(cfg, peripherals):
 
 
 def tilequeue_process_wof_neighbourhoods(cfg, peripherals):
-    from tilequeue.rawr import make_rawr_enqueuer
+    from tilequeue.stats import RawrTileEnqueueStatsHandler
     from tilequeue.wof import make_wof_model
     from tilequeue.wof import make_wof_url_neighbourhood_fetcher
     from tilequeue.wof import make_wof_processor
 
     wof_cfg = cfg.wof
     assert wof_cfg, 'Missing wof config'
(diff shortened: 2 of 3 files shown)


logging.conf.sample
```diff
@@ -1,5 +1,5 @@
 [loggers]
-keys=root,process,seed,prune_tiles_of_interest,enqueue_tiles_of_interest,dump_tiles_of_interest,load_tiles_of_interest,wof_process_neighbourhoods,query,consume_tile_traffic,tile_status,stuck_tiles,delete_stuck_tiles,rawr_enqueue,rawr_process
+keys=root,process,seed,prune_tiles_of_interest,enqueue_tiles_of_interest,dump_tiles_of_interest,load_tiles_of_interest,wof_process_neighbourhoods,query,consume_tile_traffic,tile_status,stuck_tiles,delete_stuck_tiles,rawr_enqueue,rawr_process,rawr_seed
```
zerebubuth: This is slightly OT, but; is there any way to wildcard these loggers, so that we don't have to add a new section to the config every time we use a new name?

I know in production we'd want the ability to turn them on or off, up or down individually, but it seems like it should be possible to default to `INFO` when they're not listed explicitly?

tilequeue/command.py
```diff
@@ -1628,6 +1628,47 @@ def tilequeue_rawr_process(cfg, peripherals):
...
+    group_by_zoom = rawr_yaml.get('group-zoom')
+    assert group_by_zoom is not None, 'Missing group-zoom rawr config'
+
+    rawr_queue_name = rawr_yaml.get('queue')
+    assert rawr_queue_name, 'Missing rawr queue'
+    rawr_queue = make_rawr_queue(rawr_queue_name)
+
+    msg_marshall_yaml = cfg.yml.get('message-marshall')
+    assert msg_marshall_yaml, 'Missing message-marshall config'
+    msg_marshaller = make_message_marshaller(msg_marshall_yaml)
```
zerebubuth: We seem to have this chunk of code for configuring the RAWR stuff in several places now. Is it worth moving that into `config.py`, or perhaps just to a shared function?

tilequeue/rawr.py
```diff
@@ -120,7 +120,7 @@ def __call__(self, coords):
 
         if self.logger:
             self.logger.info(
-                'Expiry processed: '
+                'Rawr tiles enqueued: '
```
zerebubuth: Well spotted! :smile:

logging.conf.sample
rmarianski: > This is slightly OT, but; is there any way to wildcard these loggers, so that we don't have to add a new section to the config every time we use a new name?

Yea, that's a good idea.

> I know in production we'd want the ability to turn them on or off, up or down individually, but it seems like it should be possible to default to INFO when they're not listed explicitly?

I think we can do it in code. What do you think of just setting the default level to INFO for all loggers, as a default parameter? Would be here:
https://github.com/tilezen/tilequeue/blob/4db723349365712d9aa6da0034196f92a285c099/tilequeue/command.py#L303

tilequeue/command.py
rmarianski: Good call. I thought it would be annoying because each part would use slightly different overlapping pieces of the rawr config, but several cases had the same boilerplate to set up the rawr enqueuer, and just factoring that out got us some good savings. Updated in 2d7f0f7.

tilequeue/rawr.py
rmarianski: It was correct at the time, in that it was only enqueueing tile from expiry lists, but now there are multiple cases :)

logging.conf.sample
rmarianski: Here's a [pr](//github.com/tilezen/tilequeue/pull/260)

PR #261, opened 2017-10-20
Emit json from rawr processing logging

rmarianski:

### DIFF
--- a/tilequeue/stats.py
+++ b/tilequeue/stats.py
@@ -56,7 +56,7 @@ class RawrTilePipelineStatsHandler(object):
                        intersect_metrics['misses'])
 
             pipe.gauge('rawr.process.enqueued', n_enqueued)
-            pipe.gauge('rawr.process.in_flight', n_enqueued)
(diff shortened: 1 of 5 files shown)


tilequeue/stats.py
```diff
@@ -56,7 +56,7 @@ def __call__(self, intersect_metrics, n_enqueued, n_inflight, timing):
                        intersect_metrics['misses'])
 
             pipe.gauge('rawr.process.enqueued', n_enqueued)
-            pipe.gauge('rawr.process.in_flight', n_enqueued)
+            pipe.gauge('rawr.process.inflight', n_inflight)
```
zerebubuth: Why the change from `rawr.process.in_flight` to `rawr.process.inflight`?

rmarianski: I just had it written as `inflight` everywhere else in the code so was just trying to be more consistent with that. In this case we won't be searching for it, but I tend to write/guess `inflight` rather than `in_flight` first in searches. But not a stronger reason other than that.

PR #267, opened 2017-10-23
Resolve new flake8 errors

rmarianski:
Connects to https://github.com/mapzen/tile-tasks/issues/302


tilequeue/command.py
```diff
@@ -457,7 +457,7 @@ def _parse_postprocess_resources(post_process_item, cfg_path):
         try:
             fn = resolve(init_fn_name)
 
-        except:
+        except Exception:
```
iandees: This feels really un-Pythonic to me, but not strongly enough to say we should ignore flake.

PR #268, opened 2017-10-24
Log coordinates as integers

rmarianski:

### DIFF
--- a/tilequeue/log.py
+++ b/tilequeue/log.py
@@ -7,9 +7,9 @@ import sys
 def make_coord_dict(coord):
     """helper function to make a dict from a coordinate for logging"""
     return dict(
-        z=coord.zoom,
-        x=coord.column,
-        y=coord.row,
+        z=int(coord.zoom),
+        x=int(coord.column),
+        y=int(coord.row),
     )
 
 
--- a/tilequeue/log.py
+++ b/tilequeue/log.py
@@ -4,12 +4,22 @@ import logging
 import sys
 
 
+def int_if_exact(x):
+    try:
+        i = int(x)
+        return i if i == x else x
+    except ValueError:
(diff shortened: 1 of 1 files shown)


zerebubuth:
Is this intended to trim trailing zeroes, e.g: print out as `{"z": 3, "x": 4, "y": 5}` rather than `{"z": 3.0, "x": 4.0, "y": 5.0}`?

I'm fairly sure that, in some circumstances, we might have fractional coordinates and it would be good to know about that when they're printed out. Perhaps we could integerise if the coordinate is within an *ulp* of being an integer already?

rmarianski:
> Is this intended to trim trailing zeroes, e.g: print out as {"z": 3, "x": 4, "y": 5} rather than {"z": 3.0, "x": 4.0, "y": 5.0}?

Yes, exactly this.

> I'm fairly sure that, in some circumstances, we might have fractional coordinates and it would be good to know about that when they're printed out. Perhaps we could integerise if the coordinate is within an ulp of being an integer already?

Really, when? Happy to just coerce to int if it's an even integer, but just curious when we would come across this scenario.

zerebubuth:
> Really, when? Happy to just coerce to int if it's an even integer, but just curious when we would come across this scenario.

The `row` and `column` of a `Coordinate` can be fractional if we zoom out and forget to call `.container()`. I can't think of an occasion where we do that deliberately, so perhaps the most likely source is a bug. But if it is a bug, and it's causing problems, then I think it would be much easier to notice if it's logged out as fractional.

rmarianski:
Ah ok, that's fair. Updated in 972c227.

rmarianski:
Assuming this is ok now and merging in. Let me know if there's some further follow up.

PR #270, opened 2017-10-24
Dockerize Tilequeue

iandees:
This adds a Dockerfile and makes some modifications so that important configuration information can be passed in via environment variables.


Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM python:2
+
+RUN apt-get -y update \
+ && apt-get -y install \
+                libgeos-dev \
+ && rm -rf /var/lib/apt/lists/*
+
+RUN mkdir -p /usr/src/app
+WORKDIR /usr/src/app
+RUN mkdir -p /usr/src/vector-datasource \
```
zerebubuth: Just wondering about how we'd control the `vector-datasource` version for dev/prod deploys, when we don't want `master`. Would we do that at image build time (perhaps by inheriting `tilequeue` and `tileserver` from the same versioned `vector-datasource` image?), or control the tag through an environment variable and have `CMD` first pull and check out that tag/branch/commit?

iandees: I know that we've gone back and forth on this a bit, but my opinion is that we always specify a specific version of in tilequeue's `requirements.txt` rather than point to a branch. This way a `pip install -r requirements.txt` always gets everything you need to run `tilequeue`. `setup.py` could have the dependencies for using `tilequeue` as a library.

zerebubuth: Ah, sorry, I didn't mean `tilequeue`, I meant `vector-datasource`. I put the comment on the wrong line, I meant to put it on the one below:

```
 && git clone --depth 1 https://github.com/mapzen/vector-datasource.git /usr/src/vector-datasource
```

PR #272, opened 2017-10-25
Handle the case where the queue returns no msgs

rmarianski:

### DIFF
--- a/tilequeue/rawr.py
+++ b/tilequeue/rawr.py
@@ -19,9 +19,10 @@ import zipfile
 
 class SqsQueue(object):
 
-    def __init__(self, sqs_client, queue_url):
+    def __init__(self, sqs_client, queue_url, recv_wait_time_seconds):
         self.sqs_client = sqs_client
         self.queue_url = queue_url
+        self.recv_wait_time_seconds = recv_wait_time_seconds
 
     def send(self, payloads):
         """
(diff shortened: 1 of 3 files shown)


tilequeue/rawr.py
```diff
@@ -249,6 +251,11 @@ def __call__(self):
                 self.log_exception(e, 'queue read')
                 continue
 
+            if not msg_handle:
+                # this gets triggered when no messages are returned
+                # TODO do we want to sleep here for a small amount of time?
```
rmarianski: I've since configured the receive wait time on the queue itself to be the maximum of 20 seconds, so at least it won't busy loop. Depending on how fast we get through diffs typically, it might make sense to sleep here for a longer time period though. Otherwise we'll just be making more sqs read calls than otherwise necessary.

iandees: It might be more obvious and less prone to accidental breakage if we pass in the receive wait param to the `receive_message()` call ([documented here](http://boto3.readthedocs.io/en/latest/reference/services/sqs.html#SQS.Client.receive_message)).

rmarianski: Yea, I thought it was actually better to push that configuration to the queue though, because if we wanted to change our minds, we just have to update the queue configuration rather than redeploy.

That being said, I don't think we'll ever change our minds in this case, and having it in the code will at least guarantee this behavior. I agree that it will make it more obvious as to what's going on as well.

Updated in ceb9f79.

PR #273, opened 2017-10-25
Add support for local storage of RAWR tiles.

zerebubuth:
Both for reading and writing. This patch means that it's possible to configure a store to round-trip RAWR tiles locally. For example, with the following RAWR config:

```yaml
rawr:
  group-zoom: 10
  queue:
    type: mem
    name: myrawrqueue
    input-file: rawr_input.txt # this just contains 10/163/395
  postgresql:
    host: null
    port: 5432
    dbname: california
    user: matt
    password: null
  store:
    type: directory
    path: rawr_tiles
  source:
    type: store
    store:
      type: directory
      name: rawr_tiles
    postgresql:
      host: null
      port: null
      dbname: california
      user: matt
    table-sources:
      planet_osm_line: &osm { name: osm, value: openstreetmap.org }
      planet_osm_point: *osm
      planet_osm_polygon: *osm
      planet_osm_ways: *osm
      planet_osm_rels: *osm
use-rawr-tiles: true
```

I'm able to run:

```
tilequeue rawr-process --config config.yaml
tilequeue tile --config config.yaml 16/10475/25320
```

And generate an empty tile. Intentionally; that tile has nothing in it - easier to check that for correctness than a few pages of GeoJSON!

### DIFF
--- a/tilequeue/command.py
+++ b/tilequeue/command.py
@@ -1198,6 +1198,7 @@ def tilequeue_process_wof_neighbourhoods(cfg, peripherals):
     from tilequeue.wof import make_wof_model
     from tilequeue.wof import make_wof_url_neighbourhood_fetcher
     from tilequeue.wof import make_wof_processor
+    from tilequeue.rawr import make_rawr_enqueuer_from_cfg
 
     wof_cfg = cfg.wof
     assert wof_cfg, 'Missing wof config'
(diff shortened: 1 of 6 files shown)


tilequeue/query/__init__.py
```diff
@@ -72,12 +72,16 @@ def _make_rawr_fetcher(cfg, layer_data, query_cfg, io_pool):
...
-    is_s3_storage = not rawr_source_yaml.get('generate-from-scratch')
-
-    if is_s3_storage:
+    # set type to "generate", and provide a postgresql subkey, to generate
+    # RAWR tiles directly, rather than trying to load them from S3. this can
+    # be useful for standalone use and testing.
+    #
+    # other types are:
+    #   s3 - to fetch RAWR tiles from S3
+    #   source - to fetch RAWR tiles from any tilequeue tile source
```
rmarianski: should this comment read `generate-from-scratch` and `store`?

tilequeue/rawr.py
```diff
@@ -446,3 +495,29 @@ def __call__(self, tile):
...
+    def _get_object(self, tile):
+        coord = unconvert_coord_object(tile)
+        format = zip_format
+        layer = 'rawr'
+        payload = self.store.read_tile(coord, format, layer)
+        return payload
+
+    def __call__(self, tile):
+        if self.io_pool:
+            payload = self.thread_pool.apply(
```
rmarianski: Should this be `self.io_pool`? And ditto for above `RawrS3Source`?

And does this return the result itself? Or an async result where we have to call `.get()` on it? And since this is just a single call, is it worth sending a single job to a separate thread only to have to immediately wait for it? Why not just do it in the current thread itself? Apologies if I missed something here.

tilequeue/command.py
```diff
@@ -1521,14 +1506,51 @@ def make_rawr_queue(name, region, wait_time_secs):
...
 
-def make_rawr_queue_from_yaml(rawr_queue_yaml):
-    name = rawr_queue_yaml.get('name')
-    assert name, 'Missing rawr queue name'
-    region = rawr_queue_yaml.get('region')
-    assert region, 'Missing rawr queue region'
-    wait_time_secs = rawr_queue_yaml.get('wait-seconds')
-    assert wait_time_secs is not None, 'Missing rawr queue wait-seconds'
-    rawr_queue = make_rawr_queue(name, region, wait_time_secs)
+class RawrMemQueue(object):
```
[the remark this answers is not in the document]
zerebubuth: Yup, thanks. Moved (plus some additional related RAWR functions) in bb13589c834f406d84932fc334d2612407715360.

tilequeue/query/__init__.py
zerebubuth: Thanks for spotting that! Yes, fixed in 1dd0c1ce40dcbdad076f2aec7f017cd23bd7779d.

tilequeue/rawr.py
zerebubuth: Good point! Removed the `io_pool` stuff in f01b007c5166a66472a91076bedbaf2ac983ba9a.

zerebubuth:
> Mind updating the sample config with what's possible now?

Of course! Done in 79fd89a16974aebd262c2e1ff3bb9b22d324d56a. As well as renaming that "mem" queue, which you rightly point out was really a "file" queue. :smile:

PR #276, opened 2017-10-27
Correct all calls to create rawr enqueuer's

rmarianski:


rmarianski:
> Ooops! Sorry I missed those. Thanks for fixing them

It might have been me :)

PR #279, opened 2017-10-30
Updates for tilequeue processing on dev

rmarianski:


tilequeue/store.py
```diff
@@ -364,6 +364,7 @@ def make_store(yml, credentials={}):
 
     elif store_type == 's3':
         bucket = yml.get('name')
+        path = yml.get('path')
```
zerebubuth: Ooops! Thanks for fixing this. I'm slightly surprised that `flake8` didn't catch it.

PR #289, opened 2017-11-08
Add command to enqueue tile pyramid

rmarianski:

### DIFF
--- a/tilequeue/command.py
+++ b/tilequeue/command.py
@@ -1,4 +1,3 @@
-from tilequeue.tile import coord_to_mercator_bounds
 from collections import defaultdict
 from collections import namedtuple
 from contextlib import closing
@@ -15,9 +14,11 @@ from tilequeue.query import DBConnectionPool
 from tilequeue.query import make_data_fetcher
 from tilequeue.queue import make_sqs_queue
 from tilequeue.store import make_store
+from tilequeue.tile import coord_children_range
 from tilequeue.tile import coord_int_zoom_up
 from tilequeue.tile import coord_is_valid
 from tilequeue.tile import coord_marshall_int
+from tilequeue.tile import coord_to_mercator_bounds
 from tilequeue.tile import coord_unmarshall_int
 from tilequeue.tile import create_coord
 from tilequeue.tile import deserialize_coord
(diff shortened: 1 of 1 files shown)


tilequeue/command.py
```diff
@@ -923,6 +924,60 @@ def tilequeue_enqueue_tiles_of_interest(cfg, peripherals):
...
+def coord_pyramids(coords, zoom_start, zoom_stop):
+    """
+    generate full pyramid for coords
+
+    Generate the full pyrmaid for the list of coords. Note that zoom_stop is
+    exclusive.
+    """
+    for coord in coords:
+        for child_coord in coord_children_range(coord, zoom_stop):
+            if zoom_start <= child_coord.zoom < zoom_stop:
```
zerebubuth: Looks like `coord_children_range` doesn't yield `coord` itself. Is that intentional?

Also, looks like the yielded coords from `coord_children_range` cannot have a zoom outside this range, so is the check here superfluous?

tilequeue/command.py
```diff
@@ -923,6 +924,60 @@ def tilequeue_enqueue_tiles_of_interest(cfg, peripherals):
...
+    assert isinstance(group_by_zoom, int), 'Invalid rawr group-zoom'
+    assert zoom_start >= group_by_zoom
+
+    ungrouped = []
+    coords_at_group_zoom = set()
+    for coord_int in tiles_of_interest:
+        coord = coord_unmarshall_int(coord_int)
+        if coord.zoom < zoom_start:
+            ungrouped.append(coord)
+        if coord.zoom >= group_by_zoom:
```
zerebubuth: Since it has been asserted that `zoom_start >= group_by_zoom`, would it be possible to spell this as `else:`?

tilequeue/command.py
rmarianski: > Looks like coord_children_range doesn't yield coord itself. Is that intentional?

Whoops, no.

> Also, looks like the yielded coords from coord_children_range cannot have a zoom outside this range, so is the check here superfluous?

I wanted to separate the zoom at which tiles were grouped, from the zoom at which jobs were emitted. In this specific case, I wanted to group at zoom level 10, but only enqueue tiles 11-15.

I think the reason I went down that path was that we had an sqs queue for z0-10, and a different one for z11+. In retrospect though, to me it's a better idea to just alter that tier to have it be z0-9 for the low zooms, and z10+ for the high.

If that's the case, then we probably want to just remove the zoom_start/zoom_stop command line args completely. The start zoom for the pyramid would just be the group by zoom from the config, and the stop zoom can be read from `cfg.max-zoom`. This will simplify the code and is probably always what we want. What do you think?

tilequeue/command.py
rmarianski: `zoom_start` and `group_by_zoom` weren't the same, but [perhaps they should be](https://github.com/tilezen/tilequeue/pull/289#discussion_r149695722).

tilequeue/command.py
zerebubuth: Ah, I get it now. The `child_coord.zoom` might be less than `start_zoom`. But I still don't see how it can be greater than or equal to `stop_zoom`.

I think it makes a lot of sense to enqueue whole pyramids. But I'm not sure that we'll always want that. Perhaps just default `start_zoom` to the group zoom and `stop_zoom` to the max zoom, but still allow them to be overridden?

tilequeue/command.py
zerebubuth: Ah, right. I see now. It's possible for a `coord` to be below `zoom_start`, but above `group_by_zoom`, when `zoom_start` is set up to render those "in between" coords individually.

Sorry about that!

tilequeue/command.py
rmarianski: > But I still don't see how it can be greater than or equal to stop_zoom.

You're right, it should never be the case. I can remove that part.

rmarianski: > Looks like coord_children_range doesn't yield coord itself. Is that intentional?

Fixed in 91804bd.

> But I still don't see how it can be greater than or equal to stop_zoom.

Fixed in b22856e.

> Perhaps just default start_zoom to the group zoom and stop_zoom to the max zoom, but still allow them to be overridden?

Good idea, updated in 216f95f.

PR #290, opened 2017-11-10
Simplify coord -> queue msg handle mappings

rmarianski:

Simplify the msg tracking logic. This should hopefully remove the errors we've been seeing. If not, instead of failing with an error, it tries to always proceed now but emit a warning in the logs for any inconsistencies.

The simplification changes:

* don't use memory ids as the indirect coord/queue handles
* use the queue message receipt handle as the key for mapping to the set of coord ids
* use a coordinate z,x,y tuple as the coord id
* use the coord_id and queue_handle_id as the coord handle, which removes the need for another intermediary mapping

The sqs queue implementation was also updated to use boto3. This was necessary to gain access to the receipt handle directly. Actually, it might have been possible with boto2 as well, but it wasn't as clear in the api, and we wanted to update to use boto3 everywhere anyway. (I thought there was an issue for this but I can't find it now).

### DIFF
--- a/tests/queue/test_msg.py
+++ b/tests/queue/test_msg.py
@@ -88,19 +88,17 @@ class SingleMessageTrackerTest(unittest.TestCase):
 
     def test_track_and_done(self):
         from tilequeue.tile import deserialize_coord
-        from tilequeue.queue import MessageHandle
-        from tilequeue.queue.message import QueueMessageHandle
-        msg_handle = MessageHandle('handle', 'payload')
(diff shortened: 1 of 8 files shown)


rmarianski:
> I notice you changed something over to boto3. Any idea how much remains using boto? Would it be worth having a task/issue to port over everything?

I thought there was an issue for it already, but I couldn't find one so I [created one](https://github.com/tilezen/tilequeue/issues/291).

PR #293, opened 2017-11-14
Updates for sqs message visibility handling

rmarianski:

There are some additional updates surrounding logging/stats here as well, as well as some fixes along the way.

### DIFF
--- a/tilequeue/command.py
+++ b/tilequeue/command.py
@@ -213,6 +213,19 @@ def make_inflight_manager(inflight_yaml, redis_client=None):
         assert 0, 'Unknown inflight type: %s' % inflight_type
 
 
+def make_sqs_queue_from_cfg(name, queue_yaml_cfg):
+    region = queue_yaml_cfg.get('region')
+    assert region, 'Missing queue sqs region'
+    visibility_extend_seconds = \
+        queue_yaml_cfg.get('visibility-extend-seconds')
+    assert visibility_extend_seconds is not None, \
+        'Missing queue sqs visibility-extend-seconds'
+    assert visibility_extend_seconds > 0, \
+        'Invalid queue sqs visibility-extend-seconds config'
+    tile_queue = make_sqs_queue(name, region, visibility_extend_seconds)
+    return tile_queue
+
+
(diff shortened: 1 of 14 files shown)


tilequeue/command.py
```diff
@@ -230,9 +259,11 @@ def make_tile_queue(queue_yaml_cfg, all_cfg, redis_client=None):
         queue_type = queue_yaml_cfg.get('type')
         assert queue_type, 'Missing queue type'
         if queue_type == 'sqs':
-            region = queue_yaml_cfg.get('region')
-            assert region, 'Missing queue region'
-            tile_queue = make_sqs_queue(queue_name, region)
+            sqs_cfg = queue_yaml_cfg.get('sqs', 'Missing queue sqs config')
```
zerebubuth: It looks like `queue_yaml_cfg` is a `dict` here, so would this assign the string `'Missing queue sqs config'` to `sqs_cfg` when the `'sqs'` entry is not present? It seems like that would give a confusing error message later. Would it be clearer to assert `sqs_cfg is not None`?

tilequeue/queue/file.py
```diff
@@ -43,6 +43,9 @@ def read(self):
     def job_done(self, msg_handle):
         pass
 
+    def job_partially_done(self, handle):
```
zerebubuth: Is there a more descriptive name for this? Perhaps `job_keepalive` or `job_made_progress`, or something that indicates that it's called repeatedly to indicate liveness?

tilequeue/queue/redis_queue.py
```diff
@@ -41,6 +41,9 @@ def read(self):
             msg_handles.append(msg_handle)
         return msg_handles
 
+    def job_paritally_done(self, handle):
```
zerebubuth: Spelling mistake in `paritally`.

tilequeue/queue/file.py
rmarianski: Good idea. I liked `job_progress` and called it that in facde1f. Sound ok?

tilequeue/queue/redis_queue.py
rmarianski: Doh! Good catch. Updated in facde1f also.

PR #298, opened 2017-11-15
Simplify _fetch_and_output function

rmarianski:

Just something I randomly noticed.


rmarianski:
Yea, I did that initially because we were just ignoring those postgresql errors anyway and accepted that we were going to get those from time to time. My thinking now is that we should see these far less frequently if we're going to be fetching most data from s3.

PR #311, opened 2017-11-28
Configurable up-zooming of tiles from Redshift

zerebubuth:
This changes the TOI gardener/pruner to parse the tile size and use the configured metatile size to up-zoom access logs to the metatile they're part of.

### DIFF
--- a/tilequeue/command.py
+++ b/tilequeue/command.py
@@ -25,6 +25,7 @@ from tilequeue.tile import coord_to_mercator_bounds
 from tilequeue.tile import coord_unmarshall_int
 from tilequeue.tile import create_coord
 from tilequeue.tile import deserialize_coord
+from tilequeue.tile import metatile_zoom_from_str
 from tilequeue.tile import parse_expired_coord_string
 from tilequeue.tile import seed_tiles
 from tilequeue.tile import serialize_coord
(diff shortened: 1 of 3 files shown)


tilequeue/command.py
```diff
@@ -1218,14 +1219,37 @@ def tilequeue_prune_tiles_of_interest(cfg, peripherals):
...
+                try:
+                    tile_size_as_zoom = metatile_zoom_from_str(tile_size)
+                    # tile size as zoom > cfg.metatile_zoom would mean that
+                    # someone requested a tile larger than the system is
+                    # currently configured to support (might have been a
+                    # previous configuration).
+                    assert tile_size_as_zoom <= cfg.metatile_zoom
+                    tile_zoom_offset = tile_size_as_zoom - cfg.metatile_zoom
+
+                except AssertionError:
```
iandees: Would this be better as an `if` block rather than a `try/assert/except`?

zerebubuth: Yes, if that were the only assertion. Unfortunately, there's also one in `metatile_zoom_from_size`, called by `metatile_zoom_from_str`, which asserts that the metatile size is a power of two. I don't think it's possible to get a (200 OK) entry in the database like that, but I think we wouldn't want this process to stop in that case.

I could rewrite both functions to return `None` on error, and check for that. Do you think that would be better than the `try/except`?

iandees: Ah, didn't see that. 👍

iandees: Ah, didn't see the other assertion. 👍

rmarianski: Catching an assertion error seems odd to me as well, and perhaps it would be better to return `None` instead and check for that at call sites.

By the way, it doesn't look like we're filtering on successful responses here, or I just missed it. Should we be?

rmarianski: By successful responses, I just meant that we'd only consider 200 and 304 responses, because we could get garbage with 404s. But, the query looks like it'll guarantee that we always get valid coordinates, so on second thought maybe it's better to just leave the way it is. After all, we are interested in the requested coordinates here, regardless of what the response looked like, eg 5xx responses should be considered.

Never mind! :)

PR #314, opened 2017-11-30
Move intersection to rawr enqueue step

rmarianski:

### DIFF
--- a/tilequeue/rawr.py
+++ b/tilequeue/rawr.py
@@ -95,21 +95,33 @@ class SqsQueue(object):
 class RawrEnqueuer(object):
     """enqueue coords from expiry grouped by parent zoom"""
 
-    def __init__(self, rawr_queue, msg_marshaller, group_by_zoom, logger,
-                 stats_handler):
+    def __init__(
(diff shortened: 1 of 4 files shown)


tilequeue/rawr.py
```diff
@@ -377,54 +402,59 @@ def __call__(self):
...
+                else:
+                    assert not is_low_zoom, \
+                        'Mix of low/high zoom coords in payload'
+
+            # check if we need to generate the rawr tile
+            # proceed directly to enqueueing the coordinates if not
+            if not is_low_zoom:
+                did_rawr_tile_gen = True
+                try:
+                    parent = common_parent(coords, self.group_by_zoom)
```
zerebubuth: I think we should also do:

```python
assert parent.zoom >= self.group_by_zoom
parent = parent.zoomTo(self.group_by_zoom).container()
```

I feel like there are going to be places in the code where we assume parent is a RAWR tile coordinate, but we'll only get a job for the partial tree and end up with `parent.zoom=11` or something like that. What do you think?

zerebubuth: Ooops... I should have looked at the definition of `common_parent`. That's exactly what it does already.

PR #320, opened 2017-12-19
Use nominal zoom in batch process.

zerebubuth:
Batch process was calling `process_coord` with the `coord.zoom` rather than the nominal zoom, meaning that all tiles were 4096 in size, and also had lower levels of detail than intended.


zerebubuth:
Screenshot of QGIS showing the buildings before this change (red) on top of those after (green). The tile is `13/2411/3079`.

image

PR #321, opened 2017-12-21
Add a backoff and retry to S3 puts to help with Slow Down messages.

zerebubuth:


tilequeue/store.py
```diff
@@ -59,6 +59,29 @@ def parse_coordinate_from_path(path, extension, layer):
...
+    from time import sleep
+
+    # do the first num_tries-1 attempts wrapped in something to catch any
+    # exceptions, optionally log them, and try again.
+    for try_counter in xrange(1, num_tries):
+        try:
+            yield
+            break
+
+        except Exception as e:
```
rmarianski: Interesting idea to use a generator, but I don't think that it'll actually catch the exception here and retry.

zerebubuth:
> Are you sure that the retry will actually catch the exception? I think it won't end up working that way.

Oooops. Wow, I hadn't realised that `yield` wasn't in the path to catch exceptions. Thanks for pointing that out.

I just pushed a different, and sadly more complex, version in 2520055. Please could you have a look and let me know if you think that's better?

tilequeue/store.py
```diff
@@ -74,12 +108,17 @@ def write_tile(self, tile_data, coord, format, layer):
...
             self.date_prefix, self.path, layer, coord, format.extension)
         key = self.bucket.new_key(key_name)
-        key.set_contents_from_string(
-            tile_data,
-            headers={'Content-Type': format.mimetype},
-            policy='public-read',
-            reduced_redundancy=self.reduced_redundancy,
-        )
+
+        @_backoff_and_retry(Exception)
```
iandees: It'd be nice if there was a logger we could pass in so we know when things retry rather than stuff just silently taking forever.

zerebubuth:
@iandees, @rmarianski does that look good with the logger threaded through?

rmarianski:
> @iandees, @rmarianski does that look good with the logger threaded through?

LGTM :+1:

PR #327, opened 2018-01-24
fix crash on using the single queue type

Akasch:
The single and multiple message per coord tracker queues are used
interchangeable on a config value, but the methods have not
the same count of arguments. This commit fixes the resulting crash
by adding the missing, unused parameter to the track function
of the SingleMessagePerCoordTracker class


rmarianski:
Thanks for the pr!

PR #330, opened 2018-04-06
Add new source for tilezen.org.

zerebubuth:
The source for the `buffered_land` changed to `tilezen.org`, which requires that tilequeue understands what that means (this is probably a design flaw / unnecessary coupling that we should fix). This PR adds that.

Connects to https://github.com/tilezen/vector-datasource/issues/1482 and https://github.com/tilezen/raw_tiles/pull/20.


nvkelso:
LGTM

nvkelso:
Followup for cleanup in https://github.com/tilezen/tilequeue/issues/331.

PR #334, opened 2018-05-03
Use boto3 for tile store

zerebubuth:
This changes the tile store implementation to use `boto3` rather than `boto` version 2. It also has a few random fixes, and extra configuration parameters.

* There's more verbose error logging for S3 errors. It now prints out the bucket and key, which should make tracking permissions errors down a bit easier.
* Use the region from the environment for Batch, if it's available. This means we're not hard-coded to `us-east-1` any more.
* Configurable object ACL. We use, `public-read` in Nextzen and this will be the default, but others may want `private`.

### DIFF
--- a/tilequeue/command.py
+++ b/tilequeue/command.py
@@ -1979,7 +1979,8 @@ def tilequeue_batch_enqueue(cfg, args):
     logger = make_logger(cfg, 'batch_enqueue')
 
     import boto3
-    client = boto3.client('batch', region_name='us-east-1')
+    region_name = os.environ.get('AWS_DEFAULT_REGION', 'us-east-1')
+    client = boto3.client('batch', region_name=region_name)
 
     logger.info('Batch enqueue ...')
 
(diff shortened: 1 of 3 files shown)


tilequeue/command.py
```diff
@@ -1979,7 +1979,8 @@ def tilequeue_batch_enqueue(cfg, args):
     logger = make_logger(cfg, 'batch_enqueue')
 
     import boto3
-    client = boto3.client('batch', region_name='us-east-1')
+    region_name = os.environ.get('AWS_DEFAULT_REGION', 'us-east-1')
+    client = boto3.client('batch', region_name=region_name)
```
rmarianski: I would assume that boto3 would automatically check the environment variable for us, but this certainly doesn't hurt.

zerebubuth: I think so too. This was my attempt to keep the default the same as before, as I think that not setting the environment variable would otherwise be an error?

rmarianski: Ah yea, good point. It does error out if there is no default region set.

👍

PR #344, opened 2018-07-12
Update s3 paths

rmarianski:
This updates the rawr and metatile s3 paths. It:

* places the hex path prefix first
* removes the unused path and layer options


rmarianski:
I added the ability to configure the s3 tile key format in 88330bb.

PR #345, opened 2018-07-13
Add metadata to rawr tile s3 PUTs

rmarianski:


tilequeue/rawr.py
```diff
@@ -580,22 +580,26 @@ class RawrS3Sink(object):
...
-                Key=location,
+        put_opts = dict(
+            Body=payload,
+            Bucket=self.bucket,
+            ContentType='application/zip',
+            ContentLength=len(payload),
+            Key=location,
         )
+        if self.metadata:
+            put_opts['Metadata'] = self.metadata
```
iandees: S3's `Metadata` is different from the `Tagging`. Check out their [docs here](https://boto3.readthedocs.io/en/latest/reference/services/s3.html#S3.Client.put_object). You want to do something like this instead:

```python
from urllib.parse import urlencode
if self.metadata:
    put_opts['Tagging'] = urlencode(self.metadata)
```

(note that boto3 `put_object()` docs say that "The tag-set must be encoded as URL Query parameters")

rmarianski: 5ed0b16

PR #347, opened 2018-07-19
Model batch jobs as aws batch array jobs

rmarianski:
This effectively unpacks the batch job types at the queue zooms (z7) one level to take advantage of batch array jobs. This involves updating the enqueue process to be set the correct job size, and modifying the rawr/low/high tile generations to be aware of this.

Unfortunately, the diff looks more aggressive than it is because the indentation changed throughout, but it boils down to removing the outer loop around what the z7 tile represents.

### DIFF
--- a/config.yaml.sample
+++ b/config.yaml.sample
@@ -439,7 +439,12 @@ batch:
   retry-attempts: 5
   # check store if metatile already exists, and if so skip processing
   check-metatile-exists: true
-  # used when enqueueing batch array jobs, can be "low" or "high"
+  # used when enqueueing batch array jobs, the job-type can be "low" or "high"
+  # this affects the batch array jobs size
+  # low:  z7 coord represents all tiles at z7, z8, z9
+  # high: z7 coord represents all tiles at z10 equivalent
+  # when enqueueing low zoom metatiles, z0-9, set to "low"
+  # when enqueueing high zoom metatiles and rawr tiles, set to "high"
   job-type: <job-type>
   # optional override memory reserved for container when submiting job
   # NOTE: in megabytes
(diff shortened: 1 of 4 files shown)


config.yaml.sample
```diff
@@ -430,6 +430,8 @@ batch:
   retry-attempts: 5
   # check store if metatile already exists, and if so skip processing
   check-metatile-exists: true
+  # used when enqueueing batch array jobs, can be "low" or "high"
```
iandees: Can you add comment about what "low" and "high" mean?

tilequeue/command.py
```diff
@@ -2116,73 +2149,95 @@ def tilequeue_meta_tile(cfg, args):
...
+        job_idx_str = os.getenv('AWS_BATCH_JOB_ARRAY_INDEX')
+        if job_idx_str is None:
+            return None
+        try:
+            job_idx = int(job_idx_str)
+        except ValueError:
+            return None
+    if job_idx < 0:
+        return None
+    return next(islice(coords, job_idx, None), None)
```
iandees: Why not do something like `coords[job_idx]`?

config.yaml.sample
rmarianski: acc3153

tilequeue/command.py
rmarianski: Doesn't work with generators unfortunately :(

```python
>>> xs = (x for x in xrange(10))
>>> xs[3]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'generator' object has no attribute '__getitem__'
```

PR #350, opened 2018-08-24
Add new coanacatl driver

zerebubuth:
It's like the existing MVT driver, but using [coanacatl](https://github.com/tilezen/coanacatl) instead of [mapbox_vector_tile](https://github.com/tilezen/mapbox_vector_tile).

It needs extra stuff installed, so it's optional at the moment. I've added a subsection to the README to call that out.

### DIFF
--- a/tilequeue/format/__init__.py
+++ b/tilequeue/format/__init__.py
@@ -3,6 +3,11 @@ from tilequeue.format.geojson import encode_single_layer as json_encode_single_l
 from tilequeue.format.mvt import encode as mvt_encode
 from tilequeue.format.topojson import encode as topojson_encode
 from tilequeue.format.vtm import merge as vtm_encode
+try:
+    from coanacatl import encode as coanacatl_encode
+except ImportError:
+    def coanacatl_encode(feature_layers, bounds_merc, extents):
+        raise RuntimeError("Could not find coanacatl library.")
 
 
 class OutputFormat(object):
@@ -62,7 +67,7 @@ def format_topojson(fp, feature_layers, zoom, bounds_merc, bounds_lnglat,
     topojson_encode(fp, features_by_layer, bounds_lnglat, extents)
 
 
-def format_mvt(fp, feature_layers, zoom, bounds_merc, bounds_lnglat, extents):
+def _make_mvt_layers(feature_layers):
     mvt_layers = []
     for feature_layer in feature_layers:
         mvt_features = []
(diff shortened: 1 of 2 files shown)


tilequeue/format/__init__.py
```diff
@@ -102,14 +119,22 @@ def format_vtm(fp, feature_layers, zoom, bounds_merc, bounds_lnglat):
 # package of tiles as a metatile zip
 zip_format = OutputFormat('ZIP Metatile', 'zip', 'application/zip',
                           None, None, None)
+# MVT, but written out by coanacatl/wagyu
+coanacatl_format = OutputFormat('MVT/Coanacatl', 'mvt',
+                                'application/x-protobuf', format_coanacatl,
+                                5, supports_shapely_geom)
```
nvkelso: Just a question: What is the "5" here?

tilequeue/format/__init__.py
```diff
@@ -102,14 +119,22 @@ def format_vtm(fp, feature_layers, zoom, bounds_merc, bounds_lnglat):
...
     topojson=topojson_format,
     vtm=vtm_format,
     mvt=mvt_format,
     mvtb=mvtb_format,
-    zip=zip_format
+    zip=zip_format,
+    # NOTE: this isn't actually the extension of the format; coanacatl writes
+    # files ending '.mvt'. this is just to give us something to call this
+    # format in config files.
+    coanacatl=coanacatl_format,
```
nvkelso: Do we need a paired change in RAWR tiles config?

tilequeue/format/__init__.py
```diff
@@ -3,6 +3,11 @@
 from tilequeue.format.mvt import encode as mvt_encode
 from tilequeue.format.topojson import encode as topojson_encode
 from tilequeue.format.vtm import merge as vtm_encode
+try:
+    from coanacatl import encode as coanacatl_encode
+except ImportError:
+    def coanacatl_encode(feature_layers, bounds_merc, extents):
+        raise RuntimeError("Could not find coanacatl library.")
```
iandees: 👍 I like the way you expose this error here.

tilequeue/format/__init__.py
zerebubuth: It's the "priority" of the driver. I'm not sure that it's used anywhere any more, but I didn't want to rip it out, just in case.

tilequeue/format/__init__.py
zerebubuth: No, it's only the tilequeue config that needs to change to support this.

PR #351, opened 2018-09-14
Add changelog and bump version to v2.1.0.

zerebubuth:


nvkelso:
@zerebubuth I reordered and formatted a few of these, and corrected links (pulls versus issues, uff). Can you have another look?

And it's not clear to me in https://github.com/tilezen/tilequeue/pull/348 if we preserved the batch job nesting as an option or tore it completely out.

Otherwise all looks good!

zerebubuth:
Looks good to me.

I ripped all the array jobs stuff out (#348) because I was seeing weird behaviour. I hope we'll be able to add it back in at some point (perhaps configurable). Unfortunately, the easiest thing to do at the time was take it all out, because it changes some of the structure of the code in ways that aren't easy to concisely switch on and off.

Issue #353, opened 2018-10-18
Support writing to multiple S3 buckets


zerebubuth:
In order to be resilient to region failure, and to get tiles closer to the clients downloading them, it would be helpful to be able to write tiles into multiple buckets.

It would be simple to support a [new type of store](https://github.com/tilezen/tilequeue/blob/44b1aa6fefe4233d6cada0fe4a903c8dbfcc3f25/tilequeue/store.py#L473-L498), perhaps called "multis3" or just switched when `name` is a list, which wraps a list of [`S3` objects](https://github.com/tilezen/tilequeue/blob/44b1aa6fefe4233d6cada0fe4a903c8dbfcc3f25/tilequeue/store.py#L121) and writes from first to last, reading from the last.

The reading from last is important so that get-before-put doesn't think that a tile written to only one of the buckets is okay. Alternatively, if we want to get more complex, we could do read repair by:

1. Reading from the first bucket - if no tile, return `None`.
2. Read from second through last buckets, if no tile then copy the one from the first bucket.
3. Return the tile.

I'm not sure whether this is worthwhile - it's a lot of extra complexity to save the work of re-rendering the tile. My feeling is that, while for some expensive tiles that would be worthwhile, the majority of tiles are so cheap to re-render that it's not worth the read repair...?

rmarianski:
> I'm not sure whether this is worthwhile

I haven't thought through the details, but my gut reaction is to just do the simplest thing, which sounds like it's just checking the get before put on the last location. IIRC we have some retries built in, ie if a write fails we will try n times before giving up. I'd imagine that this practically should cover nearly all common failures, and if there's an edge case it should be fine to re-render the odd tile. But I don't feel strongly about this, and can see the argument for optimizations too :)

PR #356, opened 2018-11-12
Write to multiple buckets.

zerebubuth:
By configuring the store bucket name as a list, this adds the ability to write to that list of buckets in order.

Connects to #353.

### DIFF
--- a/tilequeue/store.py
+++ b/tilequeue/store.py
@@ -414,14 +414,72 @@ class Memory(object):
         return [self.data] if self.data else []
 
 
+class MultiStore(object):
+    """
+    MultiStore writes to multiple stores for redundancy.
+
+    The stores are written in order from first to last, and checked in reverse
+    order. Assuming that previously-written files don't disappear (which might
+    not be true, but hopefully is at least very rare) then this should have the
+    desired behaviour under crash conditions, either:
+
+     1. The crash happened before the last tile was written, in which case some
+        of the preceding stores may not have the tile and it should be
+        re-rendered and stored again. Or,
+     2. The crash happened after the last tile was written, in which case all
+        preceding tiles should be present too.
+
+    There's an optimisation we could make later, by checking the first tile if
+    the last doesn't exist and copying it to the other stores if it does.
+    """
+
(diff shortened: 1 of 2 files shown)


tilequeue/store.py
```diff
@@ -414,14 +414,72 @@ def list_tiles(self, format):
...
                   delete_retry_interval=60, logger=None,
                   object_acl='public-read', tags=None):
     s3 = boto3.client('s3')
-    s3_store = S3(
-        s3, bucket_name, date_prefix, reduced_redundancy,
-        delete_retry_interval, logger, object_acl, tags, tile_key_gen)
+
+    # if buckets are given as a list, then write to each of them and read from
+    # the last one. this behaviour is captures in MultiStore.
+    if isinstance(bucket_name, list):
```
iandees: It might be good to make sure we still have a test that covers the single-string bucket case. I've been bitten a couple times before by the fact that Python considers a string an iterable and it acts as a list of characters. Your `isinstance(bucket_name, list)` check here will protect against that, but I could see it inadvertently getting streamlined into a iterable-ness check in the future.

tilequeue/store.py
```diff
@@ -414,14 +414,72 @@ def list_tiles(self, format):
...
                   reduced_redundancy=False, date_prefix='',
                   delete_retry_interval=60, logger=None,
                   object_acl='public-read', tags=None):
     s3 = boto3.client('s3')
-    s3_store = S3(
-        s3, bucket_name, date_prefix, reduced_redundancy,
-        delete_retry_interval, logger, object_acl, tags, tile_key_gen)
+
+    # if buckets are given as a list, then write to each of them and read from
+    # the last one. this behaviour is captures in MultiStore.
```
rmarianski: nit: captured

tilequeue/store.py
zerebubuth: Good point - added test for that in 078c26a.

tilequeue/store.py
zerebubuth: Well spotted! Fixed in d3f8277.

PR #359, opened 2018-12-05
Pass the tolerance from config through to layer datum.

zerebubuth:
The `simplify_and_clip` post-processing transform looks in the layer datum to get the tolerance factor to use for simplifying geometries. Turns out, this wasn't getting passed through from the `queries.yaml` config, so it was just using the default of `1.0` for every layer. Ooops.


zerebubuth:
It does, _if the YAML contains a float_. Otherwise it'll return whatever's in the YAML, which might cause a weird error to bubble up from the simplification step. I've added an explicit conversion to `float` in 48e18d5, just to make sure - at least the error will come from a more comprehensible layer of the stack.

PR #371, opened 2019-03-18
Bump PyPROJ version to 2.1.0, use new Transformer interface.

zerebubuth:
The internals of PyPROJ changed in v2 in a way that makes our calls to `transform` via Shapely's `ops` very slow, as PyPROJ is constantly setting up and tearing down intermediate objects. Since v2.1.0, there's a new `Transformer` interface which caches the intermediate objects, which is essentially the same speed as the old v1.9.5 version that we were using.

Since we don't need any of the new features of v2.1.0, it's tempting to just pin to the old version, however the reason I picked this up is because our tests in `vector-datasource` _don't_ pin the PyPROJ version, and we ended up getting v2, making the tests about 20x slower. Adding a pin "fixes" this, but it seems like it'd be less confusing for people using the code if the version installed by `pip install pyproj` was OK.


zerebubuth:
PS: Unit tests weren't running for this repo, looks like we never upgraded from CircleCI 1.0 and that no longer runs. I added some changes to get this working with 2.0.

zerebubuth:
PPS: This looks like it takes the `vector-datasource` CI run down from about 13-14 minutes down to 4. The time spent in the integration tests drops from 10 minutes to 40s.

PR #372, opened 2019-03-26
OpenStreetMapData.com is now hosted by FOSSGIS at osmdata.openstreetmap.de

zerebubuth:
Connects to https://github.com/tilezen/vector-datasource/issues/1855.

### DIFF
--- a/tilequeue/process.py
+++ b/tilequeue/process.py
@@ -254,8 +254,8 @@ def lookup_source(source):
         result = Source('osm', source)
     elif source == 'naturalearthdata.com':
         result = Source('ne', source)
-    elif source == 'openstreetmapdata.com':
-        result = Source('shp', source)
+    elif source in ('openstreetmapdata.com', 'osmdata.openstreetmap.de'):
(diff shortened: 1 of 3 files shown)


tilequeue/process.py
```diff
@@ -254,8 +254,8 @@ def lookup_source(source):
         result = Source('osm', source)
     elif source == 'naturalearthdata.com':
         result = Source('ne', source)
-    elif source == 'openstreetmapdata.com':
-        result = Source('shp', source)
+    elif source in ('openstreetmapdata.com', 'osmdata.openstreetmap.de'):
```
nvkelso: Why provide allowance for both here?

nvkelso: Ah, mostly for CircleCI. Should we instead make this just a test of new value, merge it and then review the main PR?

zerebubuth: It also helps to guard against using the new code with older RAWR tiles, which have the value baked-in. I think we should probably leave this in until we've done a few builds and remove it before the 1.8.0 release.

PR #374, opened 2019-04-12
Special case parse for EDTF 'uuuu', massive increase in speed. 

zerebubuth:
Also add convenience interface for writing WOF neighbourhoods SQL to file.

```
>>> import timeit
>>> timeit.timeit('parse_edtf("uuuu")', setup='from edtf import parse_edtf', number=1000)
16.14653778076172
```

At 16ms per item, parsing 90,000 neighbourhoods is going to take about 24 minutes. Having fixed this, the parsing takes about 90s.

The reason this works is that the EDTF parsing seems a little slow, but 99+% of the input is `uuuu`, so we can special-case that to a fixed value which basically takes about zero time.

### DIFF
--- a/tilequeue/wof.py
+++ b/tilequeue/wof.py
@@ -183,17 +183,22 @@ class NeighbourhoodFailure(object):
         self.superseded = superseded
 
 
-# given a string, parse it as EDTF while allowing a single 'u' or None to mean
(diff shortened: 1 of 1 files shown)


zerebubuth:
Thanks!

PR #385, opened 2021-01-17
Revert Python 3 changes

iandees:
Reverts tilezen/tilequeue#384

Although the build/tests passed, it seems that functionality is broken.


iandees:
This reverts the Python 3 changes I made, but keeps the bit of code added in that branch that skips over geometries when the intersection operation fails.

PR #388, opened 2021-05-25
python3 upgrade

peitili:


iandees:
See also https://github.com/tilezen/tilequeue/pull/384

peitili:
> See also #384 (which was reverted in #385 because it didn't actually work).

thanks, definitely a non-trivial effort.

PR #391, opened 2021-07-09
Add support to assume IAM role for accessing S3

peitili:
Add support to assume IAM role for accessing S3

At the same time remove a credentials read path since that is not used at all.

Tested manually that it works.

### DIFF
--- a/tilequeue/command.py
+++ b/tilequeue/command.py
@@ -2367,6 +2367,7 @@ def tilequeue_main(argv_args=None):
     cfg_commands = (
         ('process', tilequeue_process),
         ('seed', tilequeue_seed),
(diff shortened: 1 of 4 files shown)


tilequeue/command.py
```diff
@@ -1773,8 +1785,10 @@ def _tilequeue_rawr_setup(cfg):
 
     rawr_store = rawr_yaml.get('store')
     if rawr_store:
-        store = make_store(
-            rawr_store, credentials=cfg.subtree('aws credentials'))
```
peitili, line 1777: I removed this since the `credentials` argument is not used at all.

iandees (approved):

PR #392, opened 2021-07-10
add more verbose log for low zoom tile build

peitili:
Add more verbose log for low zoom tile build

also fix bunch of flake8 linter issues.

also add flake8 config file.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,3 @@
+[flake8]
+max-line-length = 120
+exclude = tests/*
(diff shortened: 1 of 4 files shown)


iandees (approved):
We should disable the line length QA checks or at least extend them to a sensible length (>=120 chars?)

peitili:
> We should disable the line length QA checks or at least extend them to a sensible length (>=120 chars?)

added flake8 config file to have 120 chars

PR #396, opened 2021-09-09
Add coord process total time to log

peitili:
In order for some consumer to know how much time it takes for a coord render, we need to log this information.

### DIFF
--- a/tilequeue/command.py
+++ b/tilequeue/command.py
@@ -2190,6 +2190,7 @@ def tilequeue_meta_tile(cfg, args):
             continue
 
         for fetch, coord_datum in fetched_coord_data:
+            coord_start_ms = int(time.time() * 1000)
             coord = coord_datum['coord']
             if check_metatile_exists:
                 existing_data = store.read_tile(coord, zip_format)
@@ -2234,7 +2235,8 @@ def tilequeue_meta_tile(cfg, args):
                     e, parent, job_coord, coord)
                 continue
 
-            meta_tile_logger.tile_processed(parent, job_coord, coord)
+            meta_tile_logger.tile_processed(parent, job_coord,
+                                            coord, coord_start_ms)
 
         meta_tile_logger.end_pyramid(parent, job_coord)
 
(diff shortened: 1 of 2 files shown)


tilequeue/log.py
```diff
@@ -357,6 +360,9 @@ def _log(self, msg, parent, pyramid=None, coord=None):
             json_obj['pyramid'] = make_coord_dict(pyramid)
         if coord:
             json_obj['coord'] = make_coord_dict(coord)
+        if coord_start_ms is not None:
```
travisgrigsby, line 363: do you need the `is not None`?  Looks like the case right above this doesn't use it.  No biggie either way

tilequeue/command.py
```diff
@@ -2309,6 +2311,7 @@ def tilequeue_meta_tile_low_zoom(cfg, args):
         coords.extend(coord_children_range(parent, group_by_zoom - 1))
 
     for coord in coords:
+        coord_start_ms = int(time.time() * 1000)
```
travisgrigsby, line 2314: just curious why you're making this into millis - in all places, you don't have access to the milli info, you're just multiplying by 1000, so the precision is not real;  For things under 1 second, you're going to see either 0ms or 1000ms.  Maybe just be ok with second granularity?

travisgrigsby (approved):
I think you should consider changing the duration field to be seconds not millis

tilequeue/log.py
peitili: > do you need the `is not None`? Looks like the case right above this doesn't use it. No biggie either way

since this field is integer, we also want to consider if the value is 0, that's why we cannot use the simple version

tilequeue/command.py
peitili: > just curious why you're making this into millis - in all places, you don't have access to the milli info, you're just multiplying by 1000, so the precision is not real; For things under 1 second, you're going to see either 0ms or 1000ms. Maybe just be ok with second granularity?

Graphite has precision to milliseconds and time.time() returns a float where it has some precision to the milliseconds level.

travisgrigsby: ahhh.... that's weird that Time.time returns a float, good ole python

tilequeue/log.py
travisgrigsby: sounds good

PR #397, opened 2021-09-15
Add AWS credentials refresh client helper

peitili:
Some tiles that takes a long time won't finish in one session, so we need a client which can refresh credentials automatically.


iandees (approved):
Looks good! I'm 80% sure that `boto3` has this implemented somewhere in the library, but I can't find it.

PR #400, opened 2021-10-06
add comma as character to put in quotes

travisgrigsby:
Right now, we join these hstore expressions with a "," but we don't escape strings that contain commas.   It turns out that in previous cases when we had items with commas in them, they also had spaces like 'Lincoln, Nebraska', so our hstore escaper would surround it with quotes because of the space, but we also want to do it even if there is no space.

### DIFF
--- a/tests/test_wof.py
+++ b/tests/test_wof.py
@@ -1,5 +1,7 @@
 import unittest
 
+from tilequeue.wof import escape_hstore_string
+
 
 class TestNeighbourhoodDiff(unittest.TestCase):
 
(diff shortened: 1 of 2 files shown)


iandees (approved):
Is there a test you can add for the case you found?

travisgrigsby:
> Is there a test you can add for the case you found?

Yep.  It's really easy: `name:eng_x_preferred=>a,b,c` should translate to `name:eng_x=>"a,b,c"` .

iandees (approved):

PR #401, opened 2021-10-12
Make buffer_cfg work for meta tile build

peitili:
The `buffer_cfg` exists but it doesn't work correctly, this change makes the config wired in the stack.

### DIFF
--- a/tilequeue/transform.py
+++ b/tilequeue/transform.py
@@ -106,16 +106,16 @@ def calc_max_padded_bounds(bounds, meters_per_pixel_dim, buffer_cfg):
     if buffer_cfg is None:
         return bounds
 
-    for _, format_cfg in buffer_cfg:
+    for _, format_cfg in buffer_cfg.items():
         layer_cfg = format_cfg.get('layer', {})
         if layer_cfg is not None:
-            for _, value in layer_cfg:
+            for _, value in layer_cfg.items():
                 assert isinstance(value, Number)
                 max_buffer = max(max_buffer, value)
 
         geometry_cfg = format_cfg.get('geometry', {})
         if geometry_cfg is not None:
-            for _, value in geometry_cfg:
+            for _, value in geometry_cfg.items():
                 assert isinstance(value, Number)
                 max_buffer = max(max_buffer, value)
 


tilequeue/process.py
```diff
@@ -579,14 +580,12 @@ def convert_source_data_to_feature_layers(rows, layer_data, bounds, zoom):
 
     feature_layers = []
     for layer_datum in layer_data:
+        create_query_bounds_pad_fn()
```
peitili, line 583: do you need to assign it to something?

tilequeue/query/rawr.py
```diff
@@ -724,9 +724,11 @@ def _add_feature(source, feature):
 
         return source_features.iteritems()
 
-    def __call__(self, zoom, unpadded_bounds):
+    def __call__(self, zoom, bounds):
+        """ The bounds is either an unpadded bounds if bounds_cfg is not set
+        or a padded bounds if bounds_cfg is set """
```
travisgrigsby, line 729: s/bounds_cfg/buffer_cfg

tilequeue/query/rawr.py
```diff
@@ -735,25 +737,27 @@ def __call__(self, zoom, unpadded_bounds):
         # loaded?
         assert zoom <= self.tile_pyramid.max_z
         assert zoom >= self.tile_pyramid.z
-        assert bbox.within(self.tile_pyramid.bbox())
+        # assert bbox.within(self.tile_pyramid.bbox())
```
iandees, line 740: Can this be deleted?

tilequeue/process.py
```diff
@@ -581,12 +581,9 @@ def convert_source_data_to_feature_layers(rows, layer_data, bounds, zoom):
     for layer_datum in layer_data:
         layer_name = layer_datum['name']
         features = features_by_layer[layer_name]
-        # TODO padded bounds
-        padded_bounds = dict(
-            polygon=bounds,
-            line=bounds,
-            point=bounds,
-        )
+        query_bounds_pad_fn = layer_datum['query_bounds_pad_fn']
```
travisgrigsby, line 584: @iandees this is part 2 of the change.  Rather than just using the unpadded_bounds passed in, use the query_bounds_pad_fn attached to the layer datum to create an appropriately padded bbox.

PR #402, opened 2021-12-03
Trim overzoom features from tiles where they're not needed

travisgrigsby:
In order to support overzooming, we leave features in tiles with zoom `max_zoom_with_changes` whose min_zoom is greater.  This is fine, except, because of the nature of the way we build metatiles, we also include them in the size before that, which corresponds to zoom `max_zoom_with_changes - 1`.  These features will never be used or needed in these tiles.  

This change detects this scenario, we're building a coordinate that is below `nominal_zoom` AND `nominal_zoom` equals `max_zoom_with_changes`, and then removes the features we were only leaving in to support overzoom, which is anything where `properties["min_zoom"] >= nominal_zoom +1`


nvkelso:
Reviewing output MVTs this matches my expectations.

iandees (approved):
The logic makes sense 👍🏻  but the different zooms still confuse me.

PR #404, opened 2021-12-10
make every zoom's tile use 4096 as extent

peitili:
The scale used to be 8192 in some circumstances, but we think that's not necessary so we reduce it to 4096 for all zooms.

### Test
Rebuilt England area using with the change and found all zooms extent are 4096 now.

### DIFF
--- a/tests/test_mvt.py
+++ b/tests/test_mvt.py
@@ -81,9 +81,10 @@ class MapboxVectorTileTest(unittest.TestCase):
             for (posargs, kwargs), coord in zip(encode.call_args_list,
                                                 tile_coords):
                 self.assertIn('quantize_bounds', kwargs)
-                quantize_bounds = kwargs['quantize_bounds']
-                extent = int(round((quantize_bounds[2] - quantize_bounds[0]) /
-                                   resolution))
+                # quantize_bounds = kwargs['quantize_bounds']
+                # extent = int(round((quantize_bounds[2] - quantize_bounds[0]) /
+                #                    resolution))
+                extent = 4096
                 self.assertIn('extents', kwargs)
                 actual_extent = kwargs['extents']
                 self.assertEquals(extent, actual_extent,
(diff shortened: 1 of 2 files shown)


tests/test_mvt.py
```diff
@@ -81,9 +81,10 @@ def _check_metatile(self, metatile_size):
             for (posargs, kwargs), coord in zip(encode.call_args_list,
                                                 tile_coords):
                 self.assertIn('quantize_bounds', kwargs)
-                quantize_bounds = kwargs['quantize_bounds']
-                extent = int(round((quantize_bounds[2] - quantize_bounds[0]) /
-                                   resolution))
+                # quantize_bounds = kwargs['quantize_bounds']
```
iandees, line 84: Delete these instead of commenting?

nvkelso:
In small sample tests this yields 4-5% file savings.

tilequeue/process.py
```diff
@@ -472,18 +472,16 @@ def format_coord(
...
     formatted_tiles = []
     for cut_coord in cut_coords:
-        cut_scale = _calculate_scale(scale, coord, nominal_zoom)
-
         if cut_coord == coord:
             # no need for cutting if this is the original tile.
             tiles = _format_feature_layers(
                 processed_feature_layers, coord, nominal_zoom, formats,
-                unpadded_bounds, cut_scale, buffer_cfg)
+                unpadded_bounds, scale, buffer_cfg)
```
travisgrigsby, line 479: Do you mind adding a comment here that you're not using cut scale anymore?

travisgrigsby, line 479: I think it's worth a comment that you're not using cut scale anymore

tests/test_mvt.py
peitili: > Delete these instead of commenting?

we hardcoded it to be 4096, but a better logic would be to use some of the commented out calculation; so maybe we should keep them? I added more comments/doc for this in the next commit.

tilequeue/process.py
peitili: > I think it's worth a comment that you're not using cut scale anymore

added in the next commit

iandees (approved):

nvkelso (approved):

PR #407, opened 2022-02-11
Add special handling when processing cut coord z16 tiles

peitili:
There is an issue that zoom 15, 16 mvt tiles shares a nominal zoom 16, and that nominal 16 will be used to check against the end_zoom config value in the queries.yaml file. (this issue not only happens for nominal zoom 16 but also happens for nominal zoom 0, in a word, the edge cases happens on both ends but is OK for the middle zooms) With this issue, when we write end_zoom: 17 or end_zoom:16 in the queries.yaml, both zoom 15 and 16 mvt files will be affected; if we write end_zoom: 16, only zoom 14 will be affected. So zoom15 and zoom16 mvt files have to be operated together or none.

The current tilequeue implementation only passes the nominal zoom to vector-datasource for those post_process functions defined in queries.yaml, then it cut coords to generate mvt files, it doesn’t pass the actual mvt file coord as a parameters when checking against vector-datasource. i.e. when nominal zoom is 16 it calls vector-datasource once with only 16, then it cut coords to generate zoom15 and zoom16 tiles.

One possible solution without too much refactoring I can think of is: when nominal is 16 at tilequeue, call vector-datasource twice.

1. call vector-datasource with nominal zoom 16;  cut_coords to generate zoom15 and zoom16 tiles
2. call vector-datasource with nominal zoom 17; cut_coords to generate zoom15 and zoom16 tiles.

Then select the cut zoom15 tiles from step 1 and select the cut zoom16 tiles from step 2 and concatenate them to return as the result for nominal zoom 16.

Obviously, this solution 2x the processing time for these tiles.

### DIFF
--- a/tests/test_process.py
+++ b/tests/test_process.py
@@ -258,25 +258,18 @@ class TestCalculateCutZooms(unittest.TestCase):
     def test_mid_zoom(self):
         from tilequeue.process import calculate_sizes_by_zoom
         from tilequeue.tile import metatile_zoom_from_size
-        from tilequeue.process import calculate_cut_coords_by_zoom
 
         tile_sizes = [512]
         metatile_size = 8
         metatile_zoom = metatile_zoom_from_size(metatile_size)
         max_zoom = 16 - metatile_zoom
-        print('max_zoom: {}, metatile_zoom: {}, max_zoom: {}'.format(max_zoom, metatile_zoom, max_zoom))
-        print(
-            'metatile_coord_zoom,nominal_zoom,sizes_by_zoom,'
-            'cut_coords_map_keys,cut_coords_map')
 
-        for zoom in range(0, 17):
+        for zoom in range(1, max_zoom - metatile_zoom):
             coord = Coordinate(zoom=zoom, row=0, column=0)
             sizes_by_zoom = calculate_sizes_by_zoom(
                 coord, metatile_zoom, tile_sizes, max_zoom)
             nominal_zoom = zoom + metatile_zoom
-            cut_coords_map = calculate_cut_coords_by_zoom(coord, metatile_zoom, tile_sizes, max_zoom)
-            print('{},{},"{}","{}","{}"'.format(zoom, nominal_zoom, sizes_by_zoom, cut_coords_map.keys(), cut_coords_map))
-            #self.assertEqual({nominal_zoom: tile_sizes}, sizes_by_zoom)
+            self.assertEqual({nominal_zoom: tile_sizes}, sizes_by_zoom)
 
     def test_zoom_zero(self):
         from tilequeue.process import calculate_sizes_by_zoom
--- a/tests/test_process.py
+++ b/tests/test_process.py
@@ -258,18 +258,25 @@ class TestCalculateCutZooms(unittest.TestCase):
     def test_mid_zoom(self):
         from tilequeue.process import calculate_sizes_by_zoom
         from tilequeue.tile import metatile_zoom_from_size
+        from tilequeue.process import calculate_cut_coords_by_zoom
 
         tile_sizes = [512]
         metatile_size = 8
         metatile_zoom = metatile_zoom_from_size(metatile_size)
         max_zoom = 16 - metatile_zoom
+        print('max_zoom: {}, metatile_zoom: {}, max_zoom: {}'.format(max_zoom, metatile_zoom, max_zoom))
+        print(
+            'metatile_coord_zoom,nominal_zoom,sizes_by_zoom,'
+            'cut_coords_map_keys,cut_coords_map')
 
-        for zoom in range(1, max_zoom - metatile_zoom):
+        for zoom in range(0, 17):
             coord = Coordinate(zoom=zoom, row=0, column=0)
             sizes_by_zoom = calculate_sizes_by_zoom(
                 coord, metatile_zoom, tile_sizes, max_zoom)
             nominal_zoom = zoom + metatile_zoom
-            self.assertEqual({nominal_zoom: tile_sizes}, sizes_by_zoom)
+            cut_coords_map = calculate_cut_coords_by_zoom(coord, metatile_zoom, tile_sizes, max_zoom)
+            print('{},{},"{}","{}","{}"'.format(zoom, nominal_zoom, sizes_by_zoom, cut_coords_map.keys(), cut_coords_map))
+            #self.assertEqual({nominal_zoom: tile_sizes}, sizes_by_zoom)
 
     def test_zoom_zero(self):
         from tilequeue.process import calculate_sizes_by_zoom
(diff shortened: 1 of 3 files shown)


tilequeue/process.py
```diff
@@ -556,16 +562,56 @@ def format_coord(
 # the output.
 def process_coord(coord, nominal_zoom, feature_layers, post_process_data,
                   formats, unpadded_bounds, cut_coords, buffer_cfg,
-                  output_calc_spec, scale=4096, log_fn=None, max_zoom_with_changes=16):
+                  output_calc_spec, scale=4096, log_fn=None,
```
travisgrigsby, line 565: I stared at this line for like 10 seconds before I realized there was no change

travisgrigsby: We should run our autoformatter on the whole repo

tilequeue/process.py
```diff
@@ -556,16 +562,56 @@ def format_coord(
...
 
-    return all_formatted_tiles, extra_data
+    # Because cut coord zoom 15 and 16 shares a common nominal_zoom 16,
+    # the special logic below is a necessary hack to make the
+    # current highest cut_coord zoom 16 follow the end_zoom:17 config in
+    # the post_process of queries.yaml in vector-datasource.
+    # If our highest supported zoom is not 16, the system might be
+    # broken, so we assert 16 here. Plus, we hardcoded 16 elsewhere too:
+    # https://github.com/tilezen/tilequeue/blob/43a4d4d1b101a4410660c23f1d41222e85aaa3ba/tilequeue/process.py#L382
+    assert max_zoom_with_changes == 16
```
travisgrigsby, line 583: if this works, I think we can actually do something more like
```
if max_zoom_with_changes == nominal_zoom: 
```

and then base everything off of that.  Like right now this breaks for anyone with an alternate config

tilequeue/process.py
```diff
@@ -556,16 +562,56 @@ def format_coord(
...
+        # then extract the cut coord zoom 15 tiles from the earlier formatted
+        # tiles that haven't used the hacked bumped nominal zoom 17
+        all_formatted_tiles_cut_coord_z15_z16 = \
+            [ft for ft in all_formatted_tiles if ft.get('coord').zoom == 15]
+
+        # then concatenate the result with cut coord zoom 16 tiles from the
+        # formated tiles that processed with hacked bumped nominal zoom 17
+        # as the final result
+        all_formatted_tiles_cut_coord_z15_z16.extend(
+            [ft for ft in all_formatted_tiles_special if ft.get('coord').zoom == 16])
```
travisgrigsby, line 611: does the order of these formatted tiles matter?  If not, couldn't you combine this into one comprehension?

travisgrigsby (commented):
I know we're just trying something out here - just added some comments around the direction I think we should go to support users who don't have max_zoom_with_changes==16

tests/test_process.py
```diff
@@ -258,18 +258,25 @@ def test_only_overzoom_at_max_zoom(self):
...
-        for zoom in range(1, max_zoom - metatile_zoom):
+        for zoom in range(0, 17):
             coord = Coordinate(zoom=zoom, row=0, column=0)
             sizes_by_zoom = calculate_sizes_by_zoom(
                 coord, metatile_zoom, tile_sizes, max_zoom)
             nominal_zoom = zoom + metatile_zoom
-            self.assertEqual({nominal_zoom: tile_sizes}, sizes_by_zoom)
+            cut_coords_map = calculate_cut_coords_by_zoom(coord, metatile_zoom, tile_sizes, max_zoom)
+            print('{},{},"{}","{}","{}"'.format(zoom, nominal_zoom, sizes_by_zoom, cut_coords_map.keys(), cut_coords_map))
+            #self.assertEqual({nominal_zoom: tile_sizes}, sizes_by_zoom)
```
peitili, line 279: I will revert the changes to this file

tilequeue/command.py
```diff
@@ -2187,6 +2187,8 @@ def tilequeue_meta_tile(cfg, args):
         for fetch, coord_datum in fetched_coord_data:
             coord_start_ms = int(time.time() * 1000)
             coord = coord_datum['coord']
+            # if coord.zoom != 13 and coord.row != 2866 and coord.column != 1316:
```
peitili, line 2190: all these debug comments will be reverted

tilequeue/process.py
```diff
@@ -98,6 +98,8 @@ def _log_fn(data):
             log=_log_fn,
         )
 
+        # print('peitidebug_tq_postprocess_data nominal_zoom: {}'.format(nominal_zoom))
```
peitili, line 101: all these debug comments will be reverted

tilequeue/process.py
```diff
@@ -287,6 +289,10 @@ def process_coord_no_format(
     for feature_layer in feature_layers:
         layer_datum = feature_layer['layer_datum']
         layer_name = layer_datum['name']
+
+        # if layer_name != 'buildings':
```
peitili, line 293: all these debug comments will be reverted

tilequeue/process.py
```diff
@@ -801,6 +855,7 @@ def __init__(self, coord, metatile_zoom, fetch_fn, layer_data,
         self.max_padded_bounds = calc_max_padded_bounds(self.unpadded_bounds, meters_per_pixel_dim, self.buffer_cfg)
 
     def fetch(self):
+        # metatile_zoom: 3, cfg_tile_sizes: [512], max_zoom: 13
```
peitili, line 858: all these debug comments will be reverted

tilequeue/process.py
peitili: sorry, it was an unintended change, but it does pass the linter too. anyway, I reverted the change.

tilequeue/process.py
peitili: In fact I intentionally use the hardcode value because of the comment I added above this line. 

In short, we don't know whether it shall work when max_zoom_with_changes is not 16, when it is not 16, it may not throw exception throughout the whole process, but maybe the result is logically wrong; and if this happens it is way harder to debug what's wrong since the code doesn't error out and the bug is hidden. 

Thus, we want to error out here by doing an assertion and postponing the generalization to future development. When people do want to let this handling work for non-16 value in the future, they will encounter this error here and they know what they are doing and will test and take of the result's correctness by correcting the assertion then.

peitili: In other words, I think we don't need to spend time testing the result of max_zoom_with_changes != 16 now in this PR.

tilequeue/process.py
peitili: > does the order of these formatted tiles matter? If not, couldn't you combine this into one comprehension?

I am not sure whether it matters, but  I think it is better to keep the order as before since the order is deterministic. In the for loop, lower zoom will be handled first and then higher zooms.

tilequeue/process.py
nvkelso: I'm fine keeping the hard coded 16 here, as we hard code 16 a few other places and its easier to read.

But the below section will likely significantly increase the build time, because it's doing a repeat pass in cutting the tiles. Let's see the actual impact in a build.

PR #409, opened 2022-03-07
adding a keyFormatType without a hash

travisgrigsby:
Adding a new keyFormat type `simple` that doesn't include the hash.

### DIFF
--- a/tilequeue/store.py
+++ b/tilequeue/store.py
@@ -36,6 +36,9 @@ class KeyFormatType(Enum):
     # prefix comes before hash
     prefix_hash = 2
 
+    # hash omitted
+    simple = 3
+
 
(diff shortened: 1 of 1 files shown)


tilequeue/store.py
```diff
@@ -36,6 +36,9 @@ class KeyFormatType(Enum):
     # prefix comes before hash
     prefix_hash = 2
 
+    # hash omitted
+    no_hash = 3
```
peitili, line 40: to follow the convention maybe it shall be just `prefix` ?

travisgrigsby: nm, I went with your suggestion

travisgrigsby:
ptal @peitili

peitili (approved):

PR #414, opened 2022-07-26
add support for preprocessed inline geojson layers

kevinkreiser:
just what the title says. you can now modify queries.yaml to reference layers from mercator projected geojson files. the way in which ive implemented this is fairly lightweight in that we dont support transforms or filters on the initial data (ie we dont load any yaml files like water.yaml buildings.yaml etc) because we expect the layer to be pretty much already in its final format. this does mean however that postprocessing transforms are allowed since they dont have any filtering or column juggling.

this pr is a corollary to the one over here: https://github.com/tilezen/vector-datasource/pull/2095

### DIFF
--- a/tilequeue/command.py
+++ b/tilequeue/command.py
@@ -544,6 +544,7 @@ def parse_layer_data(query_cfg, buffer_cfg, cfg_path):
             query_bounds_pad_fn=create_query_bounds_pad_fn(
                 buffer_cfg, layer_name),
             tolerance=float(layer_config.get('tolerance', 1.0)),
+            layer_path=layer_config.get('layer_path'),
         )
         layer_data.append(layer_datum)
         if layer_name in all_layer_names:
--- a/tilequeue/command.py
+++ b/tilequeue/command.py
@@ -2177,8 +2177,9 @@ def tilequeue_meta_tile(cfg, args):
         # each coord here is the unit of work now
         pyramid_coords = [job_coord]
         pyramid_coords.extend(coord_children_range(job_coord, zoom_stop))
-        # for brevity of testing it is sometimes convenient to reduce to a single child coord
-        # pyramid_coords = [Coordinate(zoom=13, column=2411, row=3080)]
+        # for brevity of testing it is sometimes convenient to reduce to a single child coord that covers your test area
+        # this makes it so that only that metatile is built rather than the whole tile pyramid which can save 20-50min
+        # pyramid_coords = [Coordinate(zoom=13, column=2411, row=3080)] # this example covers liberty island at max zoom
         coord_data = [dict(coord=x) for x in pyramid_coords]
 
         try:
--- a/tilequeue/command.py
+++ b/tilequeue/command.py
@@ -2177,6 +2177,8 @@ def tilequeue_meta_tile(cfg, args):
         # each coord here is the unit of work now
         pyramid_coords = [job_coord]
         pyramid_coords.extend(coord_children_range(job_coord, zoom_stop))
+        # for brevity of testing it is sometimes convenient to reduce to a single child coord
+        #pyramid_coords = [Coordinate(zoom=13, column=2411, row=3080)]
         coord_data = [dict(coord=x) for x in pyramid_coords]
 
         try:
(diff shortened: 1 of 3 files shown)


tilequeue/process.py
```diff
@@ -286,6 +290,12 @@ def process_coord_no_format(
     # filter, and then transform each layer as necessary
     for feature_layer in feature_layers:
         layer_datum = feature_layer['layer_datum']
+        # inline layers are expected to be pre-processed
+        layer_path = layer_datum.get('layer_path')
```
peitili, line 294: cross comment:
https://github.com/tilezen/vector-datasource/pull/2095#discussion_r930461672

tilequeue/process.py
```diff
@@ -512,6 +522,34 @@ def _calculate_scale(scale, coord, nominal_zoom):
...
 
+def _load_inline_layer(layer_path):
+    # we can optionally load geojson layers from file all we need is a file path. we expect the coordinates to already
+    # be in mercator projection and we only fill out the minimal structure of the feature tuple, ie there's no ID. if
+    # the path is given and the file exists, we expect to be able to load the layer's features. so if the json is
+    # malformed or shapely can't make sense of the geom then we'll raise
+    features = []
+
+    # has to exist and has to be geojson for now
+    if not os.path.isfile(layer_path) or not layer_path.endswith('.geojson'):
```
peitili, line 533: Is there any log we can emit to tell whether it happens, otherwise it is hard to debug if it not working as expected?

tilequeue/process.py
```diff
@@ -512,6 +522,34 @@ def _calculate_scale(scale, coord, nominal_zoom):
...
+        return features
+
+    # load the geojson into a shapely geometry collection
+    with open(layer_path) as fh:
+        fc = json.load(fh)
+        # skip if this isnt pseudo mercator
+        if fc['crs']['properties']['name'] != 'urn:ogc:def:crs:EPSG::3857':
```
peitili, line 540: Same concern, is there any log we can emit to tell whether it happens, otherwise it is hard to debug if it not working as expected?

tilequeue/process.py
```diff
@@ -512,6 +522,34 @@ def _calculate_scale(scale, coord, nominal_zoom):
...
+            return features
+        gc = GeometryCollection([shape(feature['geometry']) for feature in fc['features']])
+
+    # add the features geometries with their properties (values must be strings) in tuples
+    for geom, feat in itertools.izip(gc.geoms, fc['features']):
+        properties = {k: str(v) if type(v) is not bool else str(v).lower() for k, v in
```
peitili, line 546: maybe use another name for `v` in the `for k,v` expression to increase readability?

tilequeue/process.py
```diff
@@ -645,7 +683,7 @@ def convert_source_data_to_feature_layers(rows, layer_data, unpadded_bounds, zoo
         # expecting downstream in the process_coord function
         for layer_datum in layer_data:
             layer_name = layer_datum['name']
-            layer_props = row_props_by_layer[layer_name]
+            layer_props = row_props_by_layer.get(layer_name)
```
peitili, line 686: Any side-effect this might introduce?

tilequeue/process.py
kevinkreiser: no we specifically dont want to do that because the layer is completely optional. so it will always appear in queries.yaml but its up to the runner of the pipeline to supply the file. if they dont supply it we just ignore it

tilequeue/process.py
kevinkreiser: here we can log as it was clear the person meant to load it but screwed something up. maybe same for the file extension

tilequeue/process.py
kevinkreiser: i dont think so. if you scroll up just a bit you'll see that this object has props for every layer except this new landmarks layer. those are set to either a value or to `None`. now there isnt one for landmarks so changing this to `get` will just add handling for landmarks not existing above. alternatively i could have added landmarks above and just set it to `None` but there are a lot of places in the code that say something like "TODO: stop hardcoding all the layers" so i didnt want to add to the mess.

peitili (approved):

travisgrigsby (approved):

tilequeue/command.py
```diff
@@ -2176,6 +2177,8 @@ def tilequeue_meta_tile(cfg, args):
         # each coord here is the unit of work now
         pyramid_coords = [job_coord]
         pyramid_coords.extend(coord_children_range(job_coord, zoom_stop))
+        # for brevity of testing it is sometimes convenient to reduce to a single child coord
```
nvkelso, line 2180: Nit, say `single child cood (here over XYZ placename)`

